PR #68, opened 2021-08-10
I67 auto test

peterdudfield:
- Added github actions, to automatically run python tests
- updated conftest to 'convert_to_numpy'
- Updated geospatial so optional can be run to update the 'grids'. This was causing an error in the test. 
- Updated geospatial test to update grids first. 
- add requirements.txt file


jacobbieker (approved):
LGTM!

PR #69, opened 2021-08-15
Issue/i60 yaml settings

peterdudfield:
Want to create easy to configurations for experiments. 
Code add below put some of the basic things e.g. batchsize into a config file. 

- Created example pydantic model for configuration
- add example yaml file
- add load methods, from local and from gcs
- add save methods, from local and from gcs
- adjusted 'prepare_ml_training_data.py' script to use config file

Although pydantic models are slightly more complicated thought it worth it, for all the nice things that come with 'pydantic'


nowcasting_dataset/config/load.py
```diff
@@ -0,0 +1,59 @@
+import logging
+import gcsfs
+import os
+import io
+import yaml
+from nowcasting_dataset.config.model import Configuration
+
+logger = logging.getLogger(__name__)
```
JackKelly, line 8: We should maybe use the logger name `nowcasting_dataset`?  (Or does `__name__` resolve to `nowcasting_dataset`?!?)

nowcasting_dataset/config/load.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def load_yaml_configuration(filename: str) -> Configuration:
+    """
+    Load a yaml file which has the a configruation in it
```
JackKelly, line 13: Tiny typo fix :)   Sorry for being such a pedant!

```suggestion
    Load a yaml file which has a configuration in it
```

nowcasting_dataset/config/load.py
```diff
@@ -0,0 +1,59 @@
...
+    filename: the file name that you want to load
+    Returns: pydnaictic class
```
JackKelly, line 15:
```suggestion
    Returns: pydantic class
```

nowcasting_dataset/config/model.py
```diff
@@ -0,0 +1,52 @@
...
+class Process(BaseModel):
+    batch_size: int = Field(32, description="the batch size of the data")
+    forecast_length: int = Field(12, description="how many time steps to forecast in the future")
+    history_length: int = Field(6, description="how many historic times teps are used")
+    image_size_pixels: int = Field(64, description="the size of the satelite images")
+
+    sat_channels: tuple = Field(SAT_VARIABLE_NAMES, description="the satelite channels that are used")
+    nwp_channels: tuple = Field(NWP_VARIABLE_NAMES, description="the channels used in the nwp data")
+
+    prcesion: int = Field(16, description="what precision to use")
```
JackKelly, line 43:
```suggestion
    precision: int = Field(16, description="what precision to use")
```

nowcasting_dataset/config/save.py
```diff
@@ -0,0 +1,41 @@
...
+from nowcasting_dataset.config.model import Configuration
+
+logger = logging.getLogger(__name__)
+
+
+def save_yaml_configuration(configuration: Configuration, filename: str):
+    """
+    Load a local yaml file which has the a configuration in it
+    filename: the file name that you want to load
+    Returns: pydnaictic class
```
JackKelly, line 15:
```suggestion
    Returns: pydantic class
```

JackKelly (approved):
This looks perfect to me!  Great work!  Thank you :)

nowcasting_dataset/config/load.py
peterdudfield: no good to push me on that, my typos / spell is not very good

nowcasting_dataset/config/load.py
peterdudfield: Looks like __name__ essentially resolves it self to the file name i.e 'nowcasting_dataset.config.load'. I think this means then you can 'more' easily turn things on and off at a more granular level, if needed. It also has the bonus of if the file names change, you don't have to up date it.

This has a bit on it at the bottom
https://www.machinelearningplus.com/python/python-logging-guide/

This seems to say to do it too
https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/#:~:text=Use%20__name__%20as%20the%20logger%20name&text=as%20long%20as%20their%20logging,to%20output%20the%20logger%20name.

In a quick search in google, i cant seem to find good reasons not to do this

peterdudfield: Its funny, I just do that by habit, and I've never been asked why before

JackKelly: Cool beans, I've learnt something new, sounds like I should change to using `getLogger(__name__)`!  Thanks for looking into it!

Issue #6, opened 2021-05-07
Use CONSTS for keys into Sample dictionary


JackKelly:
e.g. so we can do `x[PV_SYSTEM_ROW_NUMBER]` instead of `x['pv_system_row_number']`

PR #73, opened 2021-08-25
Put PV strings into variables. 

peterdudfield:
Move PV strings into variables. These can then be imported in other repos. 
I couldn't see any in other data sources


peterdudfield:
issue: https://github.com/openclimatefix/nowcasting_dataset/issues/6

PR #94, opened 2021-09-07
Move tests

peterdudfield:
Move test folder


JackKelly:
Cool, sounds good!

IIRC, I put the tests below the `nowcasting_dataset/tests/` directory so stuff from one testing file could be imported into another testing file (e.g. `import nowcasting_dataset.tests.test_blah`).  But, tbh, I don't think I've _ever_ done that!  So very happy for the tests to be moved :)

PR #104, opened 2021-09-09
Add PyPi publish

jacobbieker:
# Pull Request Template

## Description

Fleshes out the setup.py a bit more, adds publishing to PyPi on release creation. As we've been doing for other OCF packages to make them easier to use, this then makes nearly all of the nowcasting things pip installable. Also added a bit to the version to make it match semantic versioning.

## How Has This Been Tested?

I built and pushed a version of the package to PyPi successfully. I then installed it with `pip install nowcasting-dataset`

Not sure of a way to test GitHub Actions other than by trying it live?

- [ ] No
- [x] Yes

## Checklist:


setup.py
```diff
@@ -1,6 +1,38 @@
...
+        "tables",
+        "boto3",
+        "moto",
+        "neptune-client",
+        "pydantic",
+        "pytest-cov",
+        "plotly",
+    ],
+    packages=find_packages(),
+)
```
peterdudfield, line 38: Should we add a long description here? So it comes up on pypi

jacobbieker: Just added it!

setup.py
```diff
@@ -1,6 +1,45 @@
...
-    version='0.1',
-    packages=find_packages())
+    name="nowcasting_dataset",
+    version="0.1.2",
+    license="MIT",
+    description="Nowcasting Dataset",
+    author="Jack Kelly, Peter Dudfield, Jacob Bieker",
+    author_email="info@openclimatefix.org",
+    company="Open Climate Fix Ltd",
+    install_requires=[
```
peterdudfield, line 16: should this be from a requirements file?

jacobbieker: Yes, it sometimes doesn't seem to work if I do it that way in Actions, and complains about the txt not existing, but I can change it back to it!

jacobbieker: Okay, turns out including a `MANIFEST.in` is the way to fix it.

peterdudfield (approved):

PR #107, opened 2021-09-10
Update NetCDFDataset to work with SatFlow

jacobbieker:
# Pull Request

## Description

This updates the NetCDFDataset to support SatFlow better. Primarily, this means adding changes to optionally ignore PV data, and adding any needed functionality from SatFlow's webdataset dataset that is missing here.

Still not the best at merging and rebasing, the changes are all under the dataset.py, although other files had formatting updated to follow OCF's black and 100 line length limit.

Now the only part of SatFlow's data pipeline that is missing is the topographic maps, but they can be added in a later PR. This should get it to working with SatFlow.

## How Has This Been Tested?

I've added two tests to the `test_netcdf_dataset.py` that check that temporal subsetting the data, and subselecting the keys we want work, and ran those! They don't run in the CI/CD on GitHub because GitHub does not have access to the data buckets.

- [] No
- [x] Yes

## Checklist:


JackKelly (commented):
Looks good so far!  :)

JackKelly:
This is very exciting stuff!  It'll be really useful when both SatFlow and predict_pv_yield can both use nowcasting_dataset.  Thanks loads for all your work on this!

JackKelly:
BTW, should I go ahead and review this PR now?  Or should I wait a bit for any more commits?  :)

jacobbieker:
> BTW, should I go ahead and review this PR now? Or should I wait a bit for any more commits? :)

I'll mark it ready for review later, and then it'll be ready! Still need to test it more, and probably add a few more things.

JackKelly:
Oops, sorry, I should've checked to see if the PR was still "work in progress"!

jacobbieker:
@JackKelly Okay, its now ready for review! It looks like a lot of changes, because rebasing after @peterdudfield's big merge seems to have added most those commits to this one... All the changes I made are contained to two files though, `dataset/datasets.py` and `test_netcdf_dataset.py`

nowcasting_dataset/dataset/datasets.py
```diff
@@ -82,19 +82,57 @@ class NetCDFDataset(torch.utils.data.Dataset):
...
-    def __init__(self, n_batches: int, src_path: str, tmp_path: str, cloud: str = "gcp"):
+
+    def __init__(
+        self,
+        n_batches: int,
+        src_path: str,
+        tmp_path: str,
+        cloud: str = "gcp",
+        required_keys: Union[Tuple[str], List[str]] = (
+            "nwp",
```
peterdudfield, line 93: Could we import these too from nowcasting_dataset.consts.py

nowcasting_dataset/dataset/datasets.py
```diff
@@ -153,26 +191,7 @@ def __getitem__(self, batch_idx: int) -> example.Example:
...
-            "pv_system_y_coords",
-            "x_meters_center",
-            "y_meters_center",
-            GSP_ID,
-            GSP_YIELD,
-            GSP_X_COORDS,
-            GSP_Y_COORDS,
-            GSP_DATETIME_INDEX,
-        ] + list(nowcasting_dataset.consts.DATETIME_FEATURE_NAMES):
+        for key in self.required_keys + list(nowcasting_dataset.consts.DATETIME_FEATURE_NAMES):
```
peterdudfield, line 194: Should we put the datetime feature into require_keys, all put a separate input as 'datetime_required_keys'

jacobbieker: I think it would make sense to put them in the required_keys, so that its all in one place.

nowcasting_dataset/dataset/datasets.py
```diff
@@ -187,6 +206,79 @@ def __getitem__(self, batch_idx: int) -> example.Example:
 
         batch = example.to_numpy(batch)
 
+        if self.current_timestep_index is not None:
```
peterdudfield, line 209: I'm suprised this is necessary - or perhaps I'm not thinking right. The batch that is loaded should have the correct timesteps in it. 

If this option used incase the .nc has a large amount of data saved like -12 hours and + 12 hours, and then this can be used to reduce it to for example -3 hours, to + 1 hour

Becasue I love testing - I would be very tempted to move this into a seperate functions and right a unittest for it

nowcasting_dataset/dataset/datasets.py
```diff
@@ -187,6 +206,79 @@ def __getitem__(self, batch_idx: int) -> example.Example:
...
+            date_time_index_to_use = (
+                "sat_datetime_index" if "sat_data" in self.required_keys else "nwp_target_time"
+            )
+            current_time = batch[date_time_index_to_use][0, self.current_timestep_index]
+            # Datetimes are in seconds, so just need to convert minutes to second + 30sec buffer
+            # Only need to do it for the first example in the batch, as masking indicies should be the same for all of them
+            start_time = current_time - (self.history_minutes * 60) - 30  # seconds
+            end_time = current_time + (self.forecast_minutes * 60) + 30  # seconds
+            if "sat_data" in self.required_keys and "sat_datetime_index" in batch:
+                sat_mask = np.logical_and(
```
peterdudfield, line 220: Can we go satellite_mask - just to be full verbose

nowcasting_dataset/dataset/datasets.py
```diff
@@ -187,6 +206,79 @@ def __getitem__(self, batch_idx: int) -> example.Example:
...
+                    start_time <= batch["sat_datetime_index"][0],
+                    batch["sat_datetime_index"][0] <= end_time,
+                )
+                _LOG.debug(f"Sat Mask: {sat_mask}")
+                batch["sat_datetime_index"] = batch["sat_datetime_index"][:, sat_mask]
+                batch["sat_data"] = batch["sat_data"][
+                    :, sat_mask, :, :, :
+                ]  # Sat is in [B, L, W, H, C]
```
peterdudfield, line 228: good comment

nowcasting_dataset/dataset/datasets.py
```diff
@@ -187,6 +206,79 @@ def __getitem__(self, batch_idx: int) -> example.Example:
...
+                ]  # PV is in [B, L, N]
+                batch["pv_elevation_angle"] = batch["pv_elevation_angle"][
+                    :, pv_mask
+                ]  # PV is in [B, L, N]
+                _LOG.debug(
+                    f"PV Datetime Shape: {batch['pv_datetime_index'].shape} PV Data Shape: {batch['pv_yield'].shape}"
+                    f" PV Azimuth Shape: {batch['pv_azimuth_angle'].shape} PV Elevation Shape: {batch['pv_elevation_angle'].shape}"
+                )
+
+            # TODO Add others for subsetting
```
peterdudfield, line 280: Do you know what other ones there are to do here?

tests/test_netcdf_dataset.py
```diff
@@ -77,3 +77,91 @@ def test_get_dataloaders_aws():
...
+
+    train_dataset.per_worker_init(1)
+    t = iter(train_dataset)
+    data = next(t)
+
+    assert "sat_data" in data.keys()
+    assert "pv_yield" not in data.keys()
+
+
+@pytest.mark.skip("CD does not have access to GCP")
```
peterdudfield, line 127: Good test - at some point we'll have to give it access to GCP. or maybe we download some .nc data and use this to run unit tests. Maybe we could make that an issue for now

tests/test_netcdf_dataset.py
```diff
@@ -77,3 +77,91 @@ def test_get_dataloaders_aws():
...
+
+    _ = torch.utils.data.DataLoader(train_dataset, **dataloader_config)
+
+    train_dataset.per_worker_init(1)
+    t = iter(train_dataset)
+    data = next(t)
+
+    sat_data = data["sat_data"]
+
+    # Sat is in 5min increments, so should have 2 history + current + 2 future
```
peterdudfield, line 165: good comments

nowcasting_dataset/dataset/datasets.py
jacobbieker: Yeah, this option is if the .nc has a large amount of data and we want to use only a subset of it, for example, some of the models that are larger might not fit in GPU memory if there is too much forecast or history timesteps, but we don't want to have to remake a second copy of the dataset just for that. I can split it into its own function and add tests though

nowcasting_dataset/dataset/datasets.py
jacobbieker: I'm not sure, I think I got them all, GSP, PV, NWP, Satellite, and datetime features, not sure if there are more to add right now

tests/test_netcdf_dataset.py
jacobbieker: We could try with some actions like these maybe? https://github.com/google-github-actions/setup-gcloud but yeah, I might add a small .nc of satellite data for unittesting the subsetting.

JackKelly (commented):
Looking great!  I'm only part way through this right now, so this is just a partial review.  I'll hopefully be able to finish off this review this afternoon :)

nowcasting_dataset/dataset/datasets.py
peterdudfield: If so, perhaps remove this comment, just to keep it tidy

nowcasting_dataset/dataset/datasets.py
jacobbieker: Added it to the required_keys

nowcasting_dataset/dataset/datasets.py
jacobbieker: I made a unit test for `subselect_data` now

nowcasting_dataset/dataset/datasets.py
```diff
@@ -82,19 +82,57 @@ class NetCDFDataset(torch.utils.data.Dataset):
...
         Args:
           n_batches: Number of batches available on disk.
           src_path: The full path (including 'gs://') to the data on
             Google Cloud storage.
           tmp_path: The full path to the local temporary directory
             (on a local filesystem).
+        required_keys: Tuple or list of keys required in the example for it to be considered usable
+        history_minutes: How many past minutes of data to use, if subsetting the batch
+        forecast_minutes: How many future minutes of data to use, if reducing the amount of forecast time
+        current_timestep_index: Index of the current timestep data, used to get the correct amount of future and past data
```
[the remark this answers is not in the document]
jacobbieker: I've updated it!

tests/test_netcdf_dataset.py
```diff
@@ -77,3 +125,91 @@ def test_get_dataloaders_aws():
...
+        os.removedirs(os.path.join(TEMP_PATH, "train"))
+    os.mkdir(os.path.join(TEMP_PATH, "train"))
+
+    train_dataset = NetCDFDataset(
+        24_900,
+        os.path.join(DATA_PATH, "train"),
+        os.path.join(TEMP_PATH, "train"),
+        cloud="gcp",
+        required_keys=[
+            "nwp",
```
peterdudfield, line 145: Can you do this from the import strings in consts.py

peterdudfield (approved):
One small change, but looks good to me

nowcasting_dataset/dataset/datasets.py
```diff
@@ -338,3 +387,93 @@ def worker_init_fn(worker_id):
...
+        forecast_minutes: How many minutes of future data to use for forecasting
+
+    Returns:
+        Example with only data between [history_minutes, forecast_minutes] remaining
+    """
+    # We are subsetting the data
+    date_time_index_to_use = (
+        SATELLITE_DATETIME_INDEX if SATELLITE_DATA in required_keys else NWP_TARGET_TIME
+    )
+    current_time = batch[date_time_index_to_use][0, current_timestep_index]
```
JackKelly, line 416: Just to be super-explicit about what's happening on this line, I'd be tempted to make use of `xarray`'s named dimensions and do something like:

```suggestion
    current_time = batch[date_time_index_to_use].isel(time=current_timestep_index)
```

(although, I must admit I haven't tested this, and I'm not 100% certain that the time dimension is called `time` in both `satellite_datetime_index` and `nwp_target_time`!)

nowcasting_dataset/dataset/datasets.py
```diff
@@ -338,3 +387,93 @@ def worker_init_fn(worker_id):
...
+    # Datetimes are in seconds, so just need to convert minutes to second + 30sec buffer
```
JackKelly, line 417: Please add a line to explain why the 30 second buffer is necessary?  (I think I can guess why, but it's always nice to be as explicit as possible!)

nowcasting_dataset/dataset/datasets.py
```diff
@@ -338,3 +387,93 @@ def worker_init_fn(worker_id):
...
+    # Only need to do it for the first example in the batch, as masking indicies should be the same for all of them
+    start_time = current_time - (history_minutes * 60) - 30  # seconds
+    end_time = current_time + (forecast_minutes * 60) + 30  # seconds
+    if SATELLITE_DATA in required_keys and SATELLITE_DATETIME_INDEX in batch:
```
JackKelly, line 421: `SATELLITE_DATETIME_INDEX` _should_ always be in the batch if `SATELLITE_DATA` is in the batch.  So I'd maybe be tempted to change this to something like:

```suggestion
    if SATELLITE_DATA in required_keys:
        assert SATELLITE_DATETIME_INDEX in batch  # Sanity check
```

(and the same goes for `if NWP_DATA in required_keys and NWP_TARGET_TIME in batch` on line 436)

nowcasting_dataset/dataset/datasets.py
```diff
@@ -338,3 +387,93 @@ def worker_init_fn(worker_id):
...
+        satellite_mask = np.logical_and(
```
JackKelly, line 422: I could be wrong but I _think_ lines 422 to 430 could be replaced with something like:

```python
for key in [SATELLITE_DATA, SATELLITE_DATETIME_INDEX]:
    batch[key] = batch[key].sel(time=slice(start_time, end_time))
```

(But, to be honest, I haven't written any `xarray` code since late-July so I definitely could be wrong!)

(If this works, then the same principle could be applied to the code below)

JackKelly (commented):
Looks good!  Just a few minor suggestions!

nowcasting_dataset/dataset/datasets.py
jacobbieker: This is currently called after the xarray data is changed to Numpy arrays, but I try moving it futher up before the call to `example.to_numpy()`

nowcasting_dataset/dataset/datasets.py
jacobbieker: Added!

nowcasting_dataset/dataset/datasets.py
jacobbieker: So, I did try doing something similar to that before, with a 2D array of start_times, end_times and it was giving errors, which is why I instead moved the subsetting code after the call to `example.to_numpy()`. I believe I might be able to do it with `.isel` if the indicies are all the same, but because each example in the batch has a different time period they cover, I don't think this would work?

nowcasting_dataset/dataset/datasets.py
JackKelly: aha, sorry, got it!  Sorry I missed that!  I guess my other comments below also don't make sense given the data is now in numpy arrays!

nowcasting_dataset/dataset/datasets.py
JackKelly: Aha, excellent point!  Sorry, I'm clearly not thinking straight!   Yeah, using `isel(time=slice(start_time, end_time))` feels like it should work (and might be less lines of code?)  But only if it makes sense to do the subsetting while the data is still in `xarray` objects

JackKelly: (I wonder if we should get into the habit of using variable names for batches like `batch_np` and `batch_xr` to make it super-explicit when we're dealing with numpy vs xarray batches??  And, yeah, I'd maybe lean towards converting to numpy as the very last step (and only because PyTorch expects numpy arrays, and doesn't know what to do with xarray objects!)

jacobbieker: It turns out that doing that doesn't work unfortunately, so not sure it'll actually be more concise, but I do agree doing it all in xarray is probably cleaner. I tried it this way: 
`current_time = batch[date_time_index_to_use].isel(time=current_timestep_index)
    start_time = current_time - pd.to_timedelta(f"{history_minutes} minute 30 second")
    end_time = current_time + pd.to_timedelta(f"{forecast_minutes} minute 30 second")
    subselect = batch[SATELLITE_DATA].isel(time=(start_time.data, end_time.data))`

jacobbieker: Got it working, and updated the code! Just need to change the test now.

JackKelly: Ah, got it... Sorry to be boring about this but please can we have one more try :)

As you probably know, `isel` expects _indicies_ into the array (just as if you were indexing into a numpy array).  For example, `isel(time=slice(4, 6))` on an xarray object is just like `array[4:6]` on a numpy array. 

So, to use `isel`, the code must first find the indicies into `time` at which `start_time` and `end_time` appear.  Finding the indicies can be done with [`np.searchsorted(a, v)`](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html) (which finds the index in `a` at which you'd insert `v` into `a` such that `a` remains sorted).

So, something vaguely like this might work (although it might be off by one!)

```python
    start_time = current_time - pd.to_timedelta(f"{history_minutes} minute 30 second")
    end_time = current_time + pd.to_timedelta(f"{forecast_minutes} minute 30 second")
    time_index_of_first_batch = batch[SATELLITE_DATA].time[0]
    start_i, end_i = np.searchsorted(time_index_of_first_batch, [start_time, end_time])
    subselect = batch[SATELLITE_DATA].isel(time=slice(start_i, end_i))
```

jacobbieker: Okay! That's a bit simpler than what I have working now, so I'll try that out!

JackKelly: Nice!  (Oops, sorry, I think we were both figuring this out at the same time - I don't think your message at 18:31 had appeared when I was writing my reply!  But well done for getting it working without my ramblings!)

jacobbieker: Had to make a few changes, but it works and is a bit more concise than what I had. The only issue left is getting the xarray for the unit test setup.

jacobbieker: Its all working now!

nowcasting_dataset/dataset/datasets.py
```diff
@@ -314,3 +364,91 @@ def worker_init_fn(worker_id):
...
+
+def subselect_data(
+    batch: example.Example,
+    required_keys: Union[Tuple[str], List[str]],
+    current_timestep_index: int,
+    history_minutes: int,
+    forecast_minutes: int,
+) -> example.Example:
+    """
+    Subselects the data temporally. This function selects all data within the range [history_minutes, forecast_minutes]
```
JackKelly, line 377: This docstring looks great!

I'm probably being overly pedantic but this tiny tweak might be worth considering!  :)

```suggestion
    Subselects the data temporally. This function selects all data within the time range [t0 - history_minutes, t0 + forecast_minutes]
```

nowcasting_dataset/dataset/datasets.py
```diff
@@ -314,3 +364,91 @@ def worker_init_fn(worker_id):
...
+
+    Args:
+        batch: Example dictionary containing at least the required_keys
+        required_keys: The required keys present in the dictionary to use
+        current_timestep_index: The index into either SATELLITE_DATETIME_INDEX or NWP_TARGET_TIME giving the current timestep
+        history_minutes: How many minutes of history to use
+        forecast_minutes: How many minutes of future data to use for forecasting
+
+    Returns:
+        Example with only data between [history_minutes, forecast_minutes] remaining
```
JackKelly, line 387: Same suggestion as above:

```suggestion
        Example with only data between [t0 - history_minutes, t0 + forecast_minutes] remaining
```

nowcasting_dataset/dataset/datasets.py
```diff
@@ -314,3 +364,91 @@ def worker_init_fn(worker_id):
...
+    """
+
+    # We are subsetting the data
+    date_time_index_to_use = (
+        SATELLITE_DATETIME_INDEX if SATELLITE_DATA in required_keys else NWP_TARGET_TIME
+    )
+    current_time = batch[date_time_index_to_use].isel(time=current_timestep_index)[0]
+    # Datetimes are in seconds, so just need to convert minutes to second + 30sec buffer
+    # Only need to do it for the first example in the batch, as masking indicies should be the same for all of them
+    # The extra 30 seconds is added to ensure that there to ensure that the first and last timestep are always contained
```
JackKelly, line 397: Thanks for adding an explanation!  A tiny typo fix:

```suggestion
    # The extra 30 seconds is added to ensure that the first and last timestep are always contained
```

nowcasting_dataset/dataset/datasets.py
```diff
@@ -314,3 +364,91 @@ def worker_init_fn(worker_id):
...
+    # within the [start_time, end_time] range
+    start_time = current_time - pd.to_timedelta(f"{history_minutes} minute 30 second")
+    end_time = current_time + pd.to_timedelta(f"{forecast_minutes} minute 30 second")
+    if SATELLITE_DATA in required_keys:
```
JackKelly, line 401: This looks good!  Thanks loads for persevering with this!

At the risk of being super-annoying, I might make one last (I promise!) suggestion to reduce the code duplication by implementing a function which gets the `time_index_of_first_example`, gets the `start_i` and `end_i`, and then uses `isel` to subset the data.

nowcasting_dataset/dataset/datasets.py
```diff
@@ -314,3 +364,91 @@ def worker_init_fn(worker_id):
...
+        time_index_of_first_batch = batch[SATELLITE_DATETIME_INDEX][0].data
```
JackKelly, line 402: Sorry, this is my mistake, I wrote this wrong in my suggestion last night!  `time_index_of_first_batch` should be `time_index_of_first_example` :)

JackKelly (approved):
Looking really good!  Thanks so much for all your work on this (and for entertaining my suggestions!)  I've made a few tiny suggestions, but nothing major!

nowcasting_dataset/dataset/datasets.py
jacobbieker: Yeah, I can add that!

PR #109, opened 2021-09-10
Issue/99 remove hash

### LINKED ISSUE #99: Remove hash from NetCDF filenames
`nowcasting_dataset` produces ML batches.  Each batch is stored in a separate NetCDF file.  The filename for each NetCDF file includes a hash.  e.g. the filename `000776_8509.nc` is made up for a hash (`000776`) and the batch number (`8509`).  I included a hash because the Google Cloud documentation says you can get slightly higher throughput by including a hash in the filename, to maximise the chance that your files are distributed across multiple storage nodes.

But the hash makes the filenames harder for humans to read.  And it's not clear that the extra performance is relevant for us.  And, we're about to do most of our ML research on our own hardware (which doesn't benefit from hashing).  So maybe we should remove the hash from the filename?

Any opinions, anyone?!

peterdudfield:
# Pull Request Template

## Description

Remove hashing from filename. 
Add optional to easily to turn it back on - this is because current dataset does include hashing in the filename, so dont wan't this change to make those files not usable. 

Fixes #99 

## How Has This Been Tested?

Unittest

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/utils.py
+++ b/nowcasting_dataset/utils.py
@@ -69,17 +69,24 @@ def sin_and_cos(df: pd.DataFrame) -> pd.DataFrame:
     return output_df
 
 
-def get_netcdf_filename(batch_idx: int) -> Path:
+def get_netcdf_filename(batch_idx: int, add_hash:bool = False) -> Path:
     """Generate full filename, excluding path.
 
     Filename includes the first 6 digits of the MD5 hash of the filename,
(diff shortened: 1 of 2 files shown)


jacobbieker (approved):
LGTM!

JackKelly (approved):
LGTM!

PR #120, opened 2021-09-14
adding .flake8 file to set line length to 100 chars

JackKelly:
# Pull Request

## Description

A two-line PR!  This one implements one way of automatically checking that lines are <= 100 chars long, as per OCF's style guide.  

## How Has This Been Tested?

This works with my code editor (Emacs).  I haven't tested it with Black yet.  It's possible Black will need another config file. 

- [ ] No
- [x] Yes


peterdudfield (approved):
Should this be put in the nowcasting meta repo - then it will be in all the repos. 
@jacobbieker @flowirtz is that how it works?

jacobbieker (approved):
LGTM! Black will do it automatically, but no harm in a second check. In https://github.com/openclimatefix/nowcasting_dataset/pull/107 I add a pre-commit hook that auto formats with Black on every push with a 100 line length limit.

JackKelly:
I'll merge for now.  We can always move to the 'meta' nowcasting repo if that's the right place for this!

Issue #115, opened 2021-09-14
Create GSP centroid


peterdudfield:
The gsp_lat and gsp_lon are at pretty arbitrary positions within each GSP region. The gsp_lat and gsp_lon define the position of the 'supergrid transformer' which serves that GSP region. The supergrid transformer could be at any arbitrary position within the region it serves (and, in some rare cases, the supergrid transformer is outside the region is serves!)

Therefore it would be good to calculate a 'centroid' position. If this takes too long, it might be good to store the metadata ourselfs, so we only have to calculate it once

PR #128, opened 2021-09-16
Issue/115 gsp centroid

peterdudfield:
# Pull Request

## Description

Get GSP centroil coordinates, and use this to make center satellite images on

Fixes issue #115 

Many changes are in nowcasting_dataset/data_sources/gsp/eso.py

## How Has This Been Tested?

Some unit tests and visualizations, check that it is in the center 

Littlebrook (_J)

Also some added notebooks, that are more like scratch files

- [ ] No
- [x] Yes

## Checklist:


JackKelly (approved):
Looks good!

One tiny request:  If we're committing the shape files into our codebase,  we should mention where the shape files came from.  Maybe in `data_sources/gsp/README.md`?  And, because the original data on the ESO website doesn't appear to have an explicit data license, we should check with Jamie and Sheffield to make sure he's happy for us to copy their data into our repo and, if he is, then we can say something like "copied with permission" in our README!  :)

peterdudfield:
> Looks good!
> 
> One tiny request: If we're committing the shape files into our codebase, we should mention where the shape files came from. Maybe in `data_sources/gsp/README.md`? And, because the original data on the ESO website doesn't appear to have an explicit data license, we should check with Jamie and Sheffield to make sure he's happy for us to copy their data into our repo and, if he is, then we can say something like "copied with permission" in our README! :)

I think he is going on holiday for two weeks, so maybe I can get round this by check the local file, and if its not there is goes and downloads it. Then we get speed ups, becasue it doesnt keep doing it and we don't commit the file

JackKelly:
That sounds good!  And that's also nice because the data might get updated by ESO, I guess.  And so users will get the updated versions of the data without us having to do anything (hopefully!)

Issue #127, opened 2021-09-16
Split: Train Test Validation 


peterdudfield:
We need a good way to split up data into train/validation/data sets. 

Would like a function that takes a a list of timestamps (`List[pd.Timestamp]`) and splits according to a few methods. Note we are dealing with timeseries data so we have to be a little careful about sampling. This is so as not to give samples that are too similar in different sets. Because Solar generation is zero in the night, we can in general make sure we split data around midnight. 

Function inputs:
- `List[pd.Timestamp]`: all time series
- `method:str`, the method to use
- `train_test_validation_split: List[int]` - This gives the ratio of data that should be put in the different sets

Methods:
- [ ] Day random - Divide the dataset into days groups, and randomly take samples according the 'train_test_validation_split'
- [ ] Week random  - Divide the dataset into weeks groups, and randomly take samples according the 'train_test_validation_split'
- [ ] Year random  - Divide the dataset into year groups, and randomly take samples according the 'train_test_validation_split'
- [ ] Specific year -  takes an input which tells which years to go into which datasets
- [ ] Full - return full dataset to all new datasets. This might be useful for unit-testing with a small amount of data
- Any others?

Bonus: 
- [ ] Might be nice to have a option to turn on random sampling, as described above, or more repetitive sample e.g 3 weeks train, 1 week validation, 1 week test and repeat. This ensures that we have data from all seasons in each dataset
- 

Return  `dict[str:List[pd.Timestamp]]` - return three datasets. The keys should be 'train','validation' and 'test'

Unittest
- [ ] This would be really great to have unit test with all these methods too

jacobbieker:
One perspective on how to split time series data is here: https://niklasriewald.com/2021/06/29/paper-review-metnet-a-neural-weather-model-for-precipitation-forecasting/

peterdudfield:
Started on https://github.com/openclimatefix/nowcasting_dataset/tree/issue/127-split-data

peterdudfield:
Decided not to do 'year random' as not sure we have enough data for this

PR #131, opened 2021-09-17
Issue/127 split data

peterdudfield:
# Pull Request

## Description

Various methods for splitting data

- same: all date goes into train, validation and test
- day repeat
- day random
- week repeat
- week random
- year specified

Fixes issue #127 

## How Has This Been Tested?

Unit tests on all methods

- [ ] No
- [x] Yes

## Checklist:


jacobbieker (approved):
Nice job, looks super clean and well done!

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+        train_test_validation_split: how the split is made
+
+    Returns: train, validation and test datetimes
+
+    """
+
+    total_weights = sum(train_test_validation_split)
+    cum_weights = np.cumsum(train_test_validation_split)
+
+    if type(datetimes) is pd.DatetimeIndex:
```
JackKelly, line 80: I could be wrong but I don't _think_ these two lines are necessary.  `pd.DatetimeIndex(datetimes)` (on line 84) will do the right thing, no matter if `datetimes` is a `List[pd.Timestamp]` or a `pd.DatetimeIndex`.

In fact, I'd recommend replacing lines 80 and 81 with: `datetimes = pd.DatetimeIndex(datetimes)` so the rest of the function knows that `datetimes` will definitely be a `DatetimeIndex`.

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+        datetimes = datetimes.to_list()
+
+    # create list of days, that are in the dataset
+    days_in_dataset = pd.DatetimeIndex(datetimes).dayofyear.unique().to_list()
```
JackKelly, line 84: Similar to my comment on `split_day()`, I think it may be safer if `days_in_dataset` were a list of _every_ unique date in the dataset, rather than a list of unique days of years.

For example, say the dataset contains 1.5 years of data, from June 2020 to Dec 2021.  Then the resulting splits will, I think, not be of the correct proportions (because there are two dates which match `dayofyear == 300`, but only one date which matches `dayofyear == 1`?

I'd recommend using something like `days_in_dataset = np.unique(datetimes.date)`

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # randomly sort indexes
+    random.shuffle(days_in_dataset)
```
JackKelly, line 87: We probably want the option to be able to seed the random number generator, so the shuffles will be identical if the input dataset is identical.

Something like this should work (assuming we pass `seed` into the function):

```python
rng = np.random.default_rng(seed)
rng.shuffle(days_in_dataset)
```

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # make which day indexes go i.e if the split is [3,1,1] then the
+    # - train_indexes = [0,1,2 ... 220,]
+    # - validation_indexes = [221, 222, ... 293]
+    # - test_indexes = [294, ...., 366]
+    train_validation_split = int(cum_weights[0] / total_weights * 365)
+    validation_test_split = int(cum_weights[1] / total_weights * 365)
+
+    train_indexes = days_in_dataset[0:train_validation_split]
```
JackKelly, line 96: I'd maybe rename these three variable from `*_indexes` to `*_dates`?  (e.g. rename `train_indexes` to `train_dates`).

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+    validation_indexes = days_in_dataset[train_validation_split:validation_test_split]
+    test_indexes = days_in_dataset[validation_test_split:]
+
+    # select the datetimes
+    train = [datetime for datetime in datetimes if datetime.dayofyear in train_indexes]
```
JackKelly, line 101: Instead of using list comprehensions, it might be worth considering using a vectorised approach, like the one I suggested in a comment on `split_day()`?  Loops might take a while if we have many years of 5-minutely data :)  But maybe this is me prematuring optimising :)

nowcasting_dataset/dataset/split/split.py
```diff
@@ -0,0 +1,80 @@
...
+    split_week_random,
+    split_week,
+)
+from enum import Enum
+
+logger = logging.getLogger(__name__)
+
+
+class SplitMethod(Enum):
+    Day = "day"
```
JackKelly, line 24: I could be wrong but should the elements on the Enum be all-caps?  e.g. replace `Day` with `DAY`?  Because they're basically consts?  But I'm really not sure - I haven't used Python Enums very much!

nowcasting_dataset/dataset/split/split.py
```diff
@@ -0,0 +1,80 @@
...
+    elif method == SplitMethod.WeekRandom:
+        train_datetimes, validation_datetimes, test_datetimes = split_week_random(
+            datetimes=datetimes, train_test_validation_split=train_test_validation_split
+        )
+    elif method == SplitMethod.Year:
+        train_datetimes, validation_datetimes, test_datetimes = split_year(
+            datetimes=datetimes, train_test_validation_year=train_test_validation_year
+        )
+    else:
+        raise
```
JackKelly, line 78: If you `raise` outside of a `try... except...` block then Python will complain `RuntimeError: No active exception to reraise` :)

Maybe replace `raise` with something like `raise ValueError('Unknown SplitMethod!')`?

nowcasting_dataset/dataset/split/week.py
```diff
@@ -0,0 +1,108 @@
...
+    # - test_indexes = [4]
+    train_indexes = [i for i in range(total_weights) if i < cum_weights[0]]
+    validation_indexes = [
+        i for i in range(total_weights) if (i >= cum_weights[0]) & (i < cum_weights[1])
+    ]
+    test_indexes = [i for i in range(total_weights) if (i >= cum_weights[1]) & (i < cum_weights[2])]
+
+    # select the datetimes
+    train = [
+        datetime for datetime in datetimes if datetime.weekofyear % total_weights in train_indexes
```
JackKelly, line 43: Similarly to the comments on `day.py`, I'd recommend filtering on _all_ the weeks in the dataset.  You can get all the unique weeks in the dataset using `unique_weeks = np.unique(datetime.to_period('W')`.

And, if you use `datetime.to_period(freq)` then that perhaps opens up the possibility of reducing code duplication by using the same function for splitting by weeks or days or months or years, by passing in `freq='D'` or `W` or `M` or `Y` :)

JackKelly (commented):
Thanks loads for your work on this!

I've made a few suggestions.  I'd definitely recommend reading through all the suggestions before implementing any of them!

In summary, I guess I'm making two main suggestions:

1. Instead of splitting based on `datetime.dayofyear` and `datetime.weekofyear`, I'd instead recommend splitting based on a list of _all_ of the unique days in the dataset, or weeks in the dataset (I explain why in some of my comments).  This can be done using `unique_periods = np.unique(datetime.to_period(freq))`.  If I've understood correctly, this would open up the possibility of using a single function to handle splitting by day, week, month or year by passing in `freq='D'` or `W` or `M` or `Y`.:)

2. Rather than using loop comprehensions to filter the datetimes, I'd recommend using a vectorised approach (which is described in one of my comments!)

Of course, these are all just suggestions!  Please push back if you think any of these suggestions are wrong or aren't worth the effort!

Thanks again for all your hard work on this!

nowcasting_dataset/dataset/split/day.py
```diff
@@ -0,0 +1,105 @@
...
+    # - test_indexes = [4]
+    train_indexes = [i for i in range(total_weights) if i < cum_weights[0]]
+    validation_indexes = [
+        i for i in range(total_weights) if (i >= cum_weights[0]) & (i < cum_weights[1])
+    ]
+    test_indexes = [i for i in range(total_weights) if (i >= cum_weights[1]) & (i < cum_weights[2])]
+
+    # select the datetimes
+    train = [
+        datetime for datetime in datetimes if datetime.dayofyear % total_weights in train_indexes
```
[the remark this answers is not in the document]
peterdudfield: Yea thats a good idea - trying to get the balance of simple code, but then not missing edge casses. 
But I think these is worth putting in

nowcasting_dataset/dataset/split/day.py
peterdudfield: Yea, in fact Ill push this higher up into the split_data function

peterdudfield: yea, ive moved this now

nowcasting_dataset/dataset/split/day.py
peterdudfield: yep. Ill switch it to using unique dates, and then just taking random dates of these

nowcasting_dataset/dataset/split/day.py
peterdudfield: hmm, i dont particualr want to, I wonder if the way to do it is, to set the seed globally, and then it would use it in these functions

nowcasting_dataset/dataset/split/day.py
peterdudfield: yea, train_dates makes more sense

nowcasting_dataset/dataset/split/day.py
peterdudfield: nah, why not, good to move over basically to pandas

nowcasting_dataset/dataset/split/split.py
peterdudfield: it appears so - https://docs.python.org/3/library/enum.html

nowcasting_dataset/dataset/split/split.py
peterdudfield: yea good point

nowcasting_dataset/dataset/split/week.py
peterdudfield: Yea, ill join it all together. Good idea with the 'freq'

nowcasting_dataset/dataset/split/day.py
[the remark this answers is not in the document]
peterdudfield: I decided not to use the inner function,  I find it sometimes hard to read - but ive written it out in a neat way - I think

peterdudfield:
@JackKelly sorry I had to move it here - https://github.com/openclimatefix/nowcasting_dataset/pull/140

My fault for being to fast on this

Issue #132, opened 2021-09-17
Make NWP data cover whole satellite image area


jacobbieker:
## Detailed Description

For SatFlow, it might be helpful to include the NWP data across the whole image. Currently, NWP is hardcoded to only get the data in the 4 pixels surrounding the center of the satellite image. Making it customizable and able to cover the whole input image gives the model a lot more information, especially for trying to determine where clouds might move, as the NWPs include wind speed, etc.

## Context

This would fix https://github.com/openclimatefix/satflow/issues/8 and would be helpful as additional input to SatFlow models.

## Possible Implementation

https://github.com/openclimatefix/nowcasting_dataset/pull/133

PR #133, opened 2021-09-17
Add customizable NWP image size

jacobbieker:
# Pull Request

## Description

Currently, NWP image size is hardcoded to 2 pixels around the PV system of interest, but for SatFlow, and potentially PV yield, we want to have the NWP data covering the whole image.

Fixes issue #132 

## How Has This Been Tested?

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/config/example.yaml
+++ b/nowcasting_dataset/config/example.yaml
@@ -14,30 +14,31 @@ process:
   batch_size: 32
   forecast_minutes: 60
   history_minutes: 30
-  image_size_pixels: 64
+  satellite_image_size_pixels: 64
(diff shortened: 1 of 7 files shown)


peterdudfield (approved):

peterdudfield:
@jacobbieker do you know why the size is 0?
https://github.com/openclimatefix/nowcasting_dataset/blob/main/tests/test_datamodule.py#L119

jacobbieker:
No, I'm not sure, changing it to be the actual NWP size, even the default of 2, fails the assertion check. So I would guess it has to do with the test Zarr file that it uses?

PR #136, opened 2021-09-17
add test data to prepare_ml_training_data script

peterdudfield:
Add test data to prepare ml training

# Pull Request

## Description

The prepared ml training data now also makes test data. 

Fixes issue #

## How Has This Been Tested?

Ran this on GCP machine

- [ ] No
- [x] Yes

## Checklist:


requirements.txt
```diff
@@ -25,4 +25,4 @@ pydantic
 pytest-cov
 plotly
 tqdm
-git+https://github.com/SheffieldSolar/PV_Live-API
+pvlive_api @ git+https://github.com/SheffieldSolar/PV_Live-API.git@master#egg=pvlive_api
```
jacobbieker, line 28: Not sure if you've had some trouble with this, but https://github.com/openclimatefix/nowcasting_dataset/issues/137 talks about how because of the git dependency, we can't upload to PyPi currently.

JackKelly (approved):
Looks good to me!

Perhaps we should also have the option to disable `test` batches (and only spit out `train` and `validation` batches?)  I'm not sure what's best when we're doing day-to-day ML research, where I've always been a bit confused about the difference between the `test` and `val` sets!?!

requirements.txt
peterdudfield: Hmm, maybe we make this an optional requirement, and that would help things

jacobbieker: Yeah, PyPi still unfortunately still won't accept it. So maybe we can make it a separate install step? But in any case, this shouldn't hold up this PR, its a different problem anyway

jacobbieker:
Ah, I was thinking we could just add it as another setup command in the workflow.yaml, before we run any of the tests

peterdudfield:
> Ah, I was thinking we could just add it as another setup command in the workflow.yaml, before we run any of the tests

yea that could be an option

peterdudfield:
> Looks good to me!
> 
> Perhaps we should also have the option to disable `test` batches (and only spit out `train` and `validation` batches?) I'm not sure what's best when we're doing day-to-day ML research, where I've always been a bit confused about the difference between the `test` and `val` sets!?!

Would this just be in the script, or would it be in the split work I was working on earlier?

One easy way to turn test data off, if to set 'n_test_batches_per_epoch' to 0

JackKelly:
> Would this just be in the script, or would it be in the split work I was working on earlier?

Ah, good question.  Without thinking about this fully, I suspect the answer is _both_ :)   That is, the user of the script should be able to turn off the test set (which, like you say, can already be done by setting `n_test_batches_per_epoch = 0`.

And then, I guess the shiny new "splitting" code would also have to be able to split the data two ways or three ways.  Although I guess the splitting code can already skip the test set if the user passes in a split like `(4, 1, 0)`?  If so, I guess we're done here, because the code already can turn off the 'test' set!

BTW, I started a new discussion 'issue' to discuss whether we _want_ a 'test' set :)  #141

PR #140, opened 2021-09-17
Issue/127 split data

peterdudfield:
# Pull Request

## Description

Update to not work on dayofyear, but look at the unique dates
Update to work on random dates in dateset, not random daysof year
Move weeks into general functions

Fixes issue #127 

## How Has This Been Tested?

unittest

- [ ] No
- [x] Yes

## Checklist:


nowcasting_dataset/dataset/split/method.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def split_method(
+    datetimes: pd.DatetimeIndex,
+    train_test_validation_split: Tuple[int] = (3, 1, 1),
+    method: str = "modulo",
+    freq: str = "D",
+) -> (List[pd.Timestamp], List[pd.Timestamp], List[pd.Timestamp]):
+    """
+    Split the data by the day into train, test and validation.
```
JackKelly, line 14:
```suggestion
    Split the data into train, test and (optionally) validation sets.
```

nowcasting_dataset/dataset/split/method.py
```diff
@@ -0,0 +1,101 @@
...
+
+    method: modulo
```
JackKelly, line 16: Very nice documentation!

nowcasting_dataset/dataset/split/method.py
```diff
@@ -0,0 +1,101 @@
...
+    If the split is (3,1,1) then, taking all the days in the dataset:
+    - train data will have all days that are modulo 5 remainder 0,1 or 2,
+    i.e 1st, 2nd, 3rd, 6th, 7th, 8th, 11th .... of the whole dataset
+    - validation data will have all days that are modulo 5 remainder 3, i.e 4th, 9th, ...
+    - test data will have all days that are modulo 5 remainder 4, i.e 5th, 10th ,  ...
+
+    method: random
+    If the split is (3,1,1) then
+    - train data will have 60% of the days
```
JackKelly, line 25: Might be worth replacing 'days' with 'data' on lines 25, 26, and 27, now that this lovely new function can handle multiple frequencies :) 

```suggestion
    - train data will have 60% of the data
```

nowcasting_dataset/dataset/split/method.py
```diff
@@ -0,0 +1,101 @@
...
+    Returns: train, validation and test datetimes
+
+    """
+
+    # find all the unique periods (dates, weeks, e.t.c)
+    datetimes_period = pd.to_datetime(datetimes.to_period(freq).to_timestamp())
+    unique_periods_in_dataset = datetimes_period.unique()
+
+    # find total weights, and cumulative weights
+    total_weights = sum(train_test_validation_split)
```
JackKelly, line 44: I think I'm just being dumb (and single-minded about what 'weights' means!) but why use the word 'weights' in the variable names `total_weights` and `cum_weights`?

nowcasting_dataset/dataset/split/method.py
```diff
@@ -0,0 +1,101 @@
...
+        train_periods = unique_periods[unique_periods["modulo"].isin(train_indexes)]["period"]
+        validation_periods = unique_periods[unique_periods["modulo"].isin(validation_indexes)][
+            "period"
+        ]
+        test_periods = unique_periods[unique_periods["modulo"].isin(test_indexes)]["period"]
+
+    elif method == "random":
+
+        # randomly sort indexes
+        unique_periods_in_dataset = np.random.permutation(unique_periods_in_dataset)
```
JackKelly, line 76: So, rather than setting the numpy random seed at the start of the data preparation script, I think it might be safer for `split_method` to construct its own private [`np.random.Generator`](https://numpy.org/doc/stable/reference/random/generator.html).  The issue with setting the global numpy random seed in the data prep script is that doesn't _guarantee_ that the splits will always be the same, because it depends on how many times `np.random` gets called before `split_method` gets called.  For example, if I wake up one morning and decide all our batches should be saved into a directory with a random name, and I call `np.random` before `split_method` gets called, then `split_method` will no longer see the same sequence of random numbers :)

So, if we passed `seed` into `split_method`, and then do something like this:

```suggestion
        rng = np.random.default_rng(seed)
        unique_periods_in_dataset = rng.permutation(unique_periods_in_dataset)
```

nowcasting_dataset/dataset/split/split.py
```diff
@@ -50,31 +46,43 @@ def split_data(
...
+    elif method == SplitMethod.WEEK_RANDOM:
+        train_datetimes, validation_datetimes, test_datetimes = split_method(
+            datetimes=datetimes,
+            train_test_validation_split=train_test_validation_split,
+            method="random",
+            freq="W",
         )
-    elif method == SplitMethod.Year:
+    elif method == SplitMethod.YEAR:
         train_datetimes, validation_datetimes, test_datetimes = split_year(
```
JackKelly, line 82: Should this line, and the following few lines, be changed to something like:

```suggestion
        train_datetimes, validation_datetimes, test_datetimes = split_method(
            datetimes=datetimes,
            train_test_validation_split=train_test_validation_split,
            method="modulo",
            freq="Y",
        )
```

JackKelly (approved):
This looks awesome!  Thank you so much for pushing away at this, and entertaining all my anal requests!  Great work!

nowcasting_dataset/dataset/split/method.py
peterdudfield: yea, becasue its adding up those weights, or ratios. Not quite sure what a better name would be, but Ill have a think

JackKelly: ah, cool, got it!  `weights` makes perfect sense now - thanks for explaining!

nowcasting_dataset/dataset/split/split.py
peterdudfield: yea, althought the year one actually takes the specific years in and splits it like that. Note like the the other methods. So there is something a bit extra in year. 

I could move this year method over to split_method

peterdudfield: actually I will do that, as the year still has a long loop in it

Issue #121, opened 2021-09-14
Get `nowcasting_dataset` running on OCF's on-premises server


JackKelly:
The source Zarr data is stored locally.

To start with, I'll copy and modify `scripts/prepare_ml_training_data.py` to create `prepare_ml_training_data_on_premises.py`, and then, depending on where we land with that, I'll maybe try to merge the changes into a single `prepare_ml_training_data.py` which works both locally and in the cloud :)

PR #142, opened 2021-09-17
Add Local Filesystem Support

jacobbieker:
# Pull Request

## Description

Allows nowcasting-dataset to read files locally, without needing to get them from GCS or AWS.

Relates to issue #121 

## How Has This Been Tested?

Unit test

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/dataset/datasets.py
+++ b/nowcasting_dataset/dataset/datasets.py
@@ -204,9 +204,12 @@ class NetCDFDataset(torch.utils.data.Dataset):
                 local_filename=local_netcdf_filename,
                 s3_resource=self.s3_resource,
             )
+        else:
+            local_netcdf_filename = remote_netcdf_filename
 
         netcdf_batch = xr.load_dataset(local_netcdf_filename)
-        os.remove(local_netcdf_filename)
+        if self.cloud != "local":
+            os.remove(local_netcdf_filename)
 
         batch = example.Example(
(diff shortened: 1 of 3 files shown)


JackKelly:
Nice!  And good timing, as we'll need this next week to get the code running on OCF ML server :)

nowcasting_dataset/dataset/datasets.py
```diff
@@ -159,15 +159,15 @@ def __init__(
         self.gcs = None
         self.s3_resource = None
 
-        assert cloud in ["gcp", "aws"]
+        assert cloud in ["gcp", "aws", "local"]
```
JackKelly, line 162: I wonder if we should rename `cloud` to `compute_environment` or something like that?  `cloud = 'local'` feels a bit weird.  I don't have a strong opinion though!

JackKelly (approved):
Looks good to me!  (Although I admit I haven't yet fully wrapped my head around what's required to get `nowcasting_dataset` running locally!)

nowcasting_dataset/dataset/datasets.py
jacobbieker: If we are going for being explicit, then probably should be `compute_environment`, but our own stuff is still just our own local `cloud` too!

JackKelly: Yeah, true!  So, yeah, I really don't mind what we do here.  Very happy to stick with `cloud`!

Issue #143, opened 2021-09-20
GSP variables


peterdudfield:
**Describe the bug**
'gsp_datetime_index' is not in prepared ML data

**To Reproduce**
Use test_netcdf_dataset.py and load from "gs://solar-pv-nowcasting-data/prepared_ML_training_data/v6/"

**Expected behavior**
'gsp_datetime_index' to be in dataset

PR #145, opened 2021-09-20
Bug/143 gsp datetime

peterdudfield:
# Pull Request

## Description

Add gsp_datatime_index to the prepared ml data. 

Fixes issue #143 

## How Has This Been Tested?

- Unittests 
- Run machine learning prepare ml script and load data and check that it the variable was there

- [ ] No
- [x] Yes

## Checklist:


nowcasting_dataset/dataset/batch.py
```diff
@@ -101,9 +102,14 @@ def batch_to_dataset(batch: List[Example]) -> xr.Dataset:
 
             # GSP
             n_gsp_systems = len(example[GSP_ID])
-            one_dateset["gsp_yield"] = xr.DataArray(
+            one_dateset[GSP_YIELD] = xr.DataArray(
```
JackKelly, line 105: I'm now about to show my own confusion about `xarray` :)   This comment is as much a question as a code review suggestion!

Is `example[GSP_YIELD]` a `pd.DataFrame` with a `pd.DatetimeIndex` as its index?

If so, then I think the coords for the `time_30` dimension in `one_dateset[GSP_YIELD]` will be an array of _datetimes_ and so won't align with the array of _integer_ coords for the `time_30` dimension in `one_dataset[GSP_DATETIME_INDEX]`.  If that makes sense?!?  (I had to refer to [`xarray`'s documentation](http://xarray.pydata.org/en/stable/user-guide/terminology.html) to remind myself of the distinction between coords and dimensions!)

On the other hand, if `example[GSP_YIELD]` is a numpy array (or a `pd.DataFrame` with an integer index) then all is fine!

Which also makes me think that I should've got into the habit of commenting exactly what type of object each field in `example` is :)

JackKelly (approved):
Looks good to me!  Thanks for your work on this :)

Just two tiny comments!

nowcasting_dataset/dataset/batch.py
peterdudfield: Its a numpy array, so I'll add that into the documentation

Issue #154, opened 2021-09-22
"UserWarning: Geometry is in a geographic CRS. Results from 'centroid' are likely incorrect."


JackKelly:
**Describe the bug**
When running `prepare_ml_data.py` I get this warning... 

> /home/jack/dev/ocf/nowcasting_dataset/nowcasting_dataset/data_sources/gsp/eso.py:67: UserWarning: Geometry is in a geographic CRS. Results from 'centroid' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.

If we're sure this is a false warning then let's silence the warning :)

peterdudfield:
Yea I found this strange, as I had already used a function to move this to 'crs'. We could hind the comment in the script. I better way would be to open up a bug report on geopandas

JackKelly:
I must admit I had to Google what a "geographic CRS" is :)

[It turns out that](https://www.earthdatascience.org/courses/use-data-open-source-python/intro-vector-data-python/spatial-data-vector-shapefiles/geographic-vs-projected-coordinate-reference-systems-python/) a _geographic_ CRS is a coordinate reference system which spans the entire globe (such as latitude & longitude) and hence might suffer from some local distortions (because Earth isn't a perfect sphere).  In contrast, a _projected_ CRS is one that only covers a bit of the globe, and hence can be better optimised to that region.

So I'd _guess_ this warning is saying that the centroid may be wrong by something like a few tens of meters because lat / lon is geographic CRS (and hence is slightly distorted).

TBH, when we're computing the centroid of each GSP region, we probably don't care about super-high accuracy:  We probably only need the centroid to be accurate to within a kilometre or so.  So I'd guess we can safely ignore this warning :)

@peterdudfield I remember you plotted some maps when you did this work originally.  Did your plots include the computed centroid super-imposed on the shape of the GSP?  And did the centroids look about right?!  If so, I guess we can be pretty confident that we can ignore this warning?

One solution maybe be to simply leave the shape file from ESO in its native [OSGB projection](https://epsg.io/27700) (instead of converting to lat/lon... i.e. on [line 108](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/data_sources/gsp/eso.py#L108), we'd remove `.to_crs(WGS84_CRS)`)?  OSGB is a "projected CRS".  [WGS84_CRS is another name for lat/lon](https://epsg.io/4326).

[...]

PR #165, opened 2021-09-27
calculate centroid before using to_crs()

peterdudfield:
# Pull Request

## Description

Removed warning about
"UserWarning: Geometry is in a geographic CRS. Results from 'centroid' are likely incorrect."

Fixes issue #154 

## How Has This Been Tested?

- normal unittests
- added small change to unitest
- run ploting script

- [ ] No
- [x] Yes

## Checklist:


peterdudfield:
ok to merge this @JackKelly ?

nowcasting_dataset/data_sources/gsp/eso.py
```diff
@@ -21,10 +21,19 @@
 import geopandas as gpd
 import pandas as pd
 
-from nowcasting_dataset.geospatial import WGS84_CRS
+from nowcasting_dataset.geospatial import osgb_to_lat_lon, WGS84_CRS
+
 
 logger = logging.getLogger(__name__)
 
+rename_save_columns = {
```
JackKelly, line 29: Maybe in another PR, we could make it so we don't have to rename the columns, but instead we use the same column names throughout the code?

nowcasting_dataset/data_sources/gsp/eso.py
```diff
@@ -105,10 +112,26 @@ def get_gsp_shape_from_eso(
...
+
+            # calculate the centroid before using - to_crs
+            shape_gpd["centroid_x"] = shape_gpd["geometry"].centroid.x
+            shape_gpd["centroid_y"] = shape_gpd["geometry"].centroid.y
+            shape_gpd["centroid_lat"], shape_gpd["centroid_lon"] = osgb_to_lat_lon(
+                x=shape_gpd["centroid_x"], y=shape_gpd["centroid_y"]
+            )
+
+            # project to WGS84
+            shape_gpd = shape_gpd.to_crs(WGS84_CRS)
```
JackKelly, line 125: Is this still necessary?  I think the satellite data and the NWPs are in OSGB projection in the batches, right?  It'd be good to keep all the geospatial data in the same projection in the batches.  Or have I misunderstood?!

tests/data_sources/gsp/test_gsp_data_source.py
```diff
@@ -46,6 +41,13 @@ def test_gsp_pv_data_source_get_locations_for_batch():
     locations_x, locations_y = gsp.get_locations_for_batch(t0_datetimes=gsp.gsp_power.index[0:10])
 
     assert len(locations_x) == len(locations_y)
+    assert locations_x[0] > 90  # this makes sure it is not in lat/lon
```
JackKelly, line 44: OSGB coordinates _can_ be less than or equal to 90 :)  I'm not sure of any perfect way to assert that a coordinate is OSGB or lat,lon.  Other than, maybe, checking that it's within a certain geospatial boundary (e.g. is it on land within the UK?)  But that's maybe overkill?!

JackKelly (approved):
Looks good to me!  Just a few quick thoughts.  Sorry for taking a while to review this!

nowcasting_dataset/data_sources/gsp/eso.py
peterdudfield: Yea I did this as the column names seemed to be limited to 10, but let me check this again

tests/data_sources/gsp/test_gsp_data_source.py
peterdudfield: hmm, good point, I think if it was less than 90 it would be in the sea - and we shouldnt have a GSP centroid there, so I'm happy to keep it like this - perhaps Ill add a comment in the code?

nowcasting_dataset/data_sources/gsp/eso.py
peterdudfield: https://en.wikipedia.org/wiki/Shapefile#Limitations and https://github.com/geopandas/geopandas/issues/1417
says its limited to 10

peterdudfield: And then its a balance of a bit of extra code here to give full verbose, compared to a shorter column name and it not being as verbose. Personally I like the fully verbose version

tests/data_sources/gsp/test_gsp_data_source.py
JackKelly: Sure, sounds good:  Leave the code as-it but add a comment to explain...

nowcasting_dataset/data_sources/gsp/eso.py
JackKelly: Ah, sorry, my mistake... I hadn't realised shapefiles have these limitations!  Thanks for explaining.  Cool, let's leave the code as-is!

PR #167, opened 2021-09-27
Origin/peter/on premises

peterdudfield:
change requirement order
- not sure why this makes a difference
- mite be something to do with boto and moto


JackKelly:
Thanks loads for looking into this @peterdudfield !

Issue #91, opened 2021-09-03
Implement script to download data from EUMETSAT Data Store


JackKelly:
https://www.eumetsat.int/eumetsat-data-store

Required behaviour when the script runs:

1. Check what data is already downloaded to disk (e.g. maybe all data for 2020-01 and 2020-06 is already downloaded, so we don't need to download those two months of data again.)
2. If the script is called with the `--backfill` flag then download any missing data between the start date of the data available from EUMETSAT and the end-date of the downloaded data.  (e.g. download 2010 to 2019-12-31; and download 2020-02-01 to 2020-05-30)
3. If the `--backfill` flag is not set then just download any missing new data.

For now, we just want the EUMETSAT SEVIRI RSS data (full spatial extent, and all timesteps).  In the future, we may want other EUMETSAT data products.

Download concurrently using multiple threads.

While the file is downloading, append `partial` to the filename.  Only remove the `partial` once the file has fully downloaded and passes all sanity checks (so, if the script crashes part way through a download, it's clear which files are partially downloaded and should be deleted!)

Log download progress.  

If possible, allow the user to limit the download rate (so I can still watch Netflix whilst downloading EUMETSAT data!)

Don't bother with spatial re-projection yet (that will be done when nowcasting_dataset is run).

@jacobbieker is your EUMETSAT data ingestion script on GitHub?  (No worries if not!)  If it is, please could you point us to it?

Maybe _don't_ compress as `.bz2` because that forces us to load the entire `.bz2` file off disk and decompress it, even if we only want a small chunk.

Related issues:

* #95 
* #81

jacobbieker:
Its here: https://github.com/openclimatefix/satflow/blob/2be1708bb37916d23ffcf18dd1a1cfd6a709acf5/satflow/data/download.py#L179-L206 using a slightly modified version of https://github.com/openclimatefix/Satip/tree/bug-fix That code downloads the binary cloud mask product as well

JackKelly:
Hi @jacobbieker do you have any plans for how the script will store the raw EUMETSAT data on disk?  Maybe a simple directory structure like `<year>/<month>/<day>/<files>`?  I have no strong feelings either way!

I had previously used `<year>/<month>/<day>/<hour>/<minute of data acquisition>` but I think that's too many levels!  And I had previous compressed the raw files as `bzip2` but I now think that's a bad idea because that forces us to load and decompress the entire file, even if we only want a small subset of the file.

[...]

Issue #95, opened 2021-09-07
Forwards-fill training dataset (i.e. get most recent data)


JackKelly:
This is a sub-task of #81

JackKelly:
I think this is done now, so I'll close this issue for now.  Please re-open if I'm mistaken!

Issue #81, opened 2021-09-01
Get data from 2020-01-01 to 2021-10-01


JackKelly:
We're currently using less than 2 years of data.  But about 10 years of data is available!  More training data is usually a Good Thing :)

The priority is probably to forward-fill our training dataset (because the PV Live data & ESO's PV forecasts are both quite recent).

## Tasks
- [x] #95.  This is probably the priority
- [x] Back-fill data (i.e. extend the training dataset backwards in time).
- [x] Add more disks to OCF's ML server (currently has 8 x 12 TB disks in a ZFS RAIDZ array)

## Types of data
- [x] EUMETSAT SEVIRI RSS (2020 and 2021 is downloaded and converted to Zarr.)
- [x] Numerical Weather Predictions  (downloaded all data from 2016 to 2021.  Converted 2020 and 2021 to Zarr)
- [x] PV time series
- [x] GSP PV data from Sheffield Solar

jacobbieker:
Just to note, my code works, but sometimes hangs and has to be restarted. I would think Airflow pipeline might be better suited for continually ingesting data?

JackKelly:
Cool beans!

@AyrtonB what would you recommend for ingesting ~10 years of EUMETSAT data onto our "in-premises" Linux server?  I remember you were looking at a more recent alternative to Airflow (DAGster?)  Would you recommend we use custom code; or Airflow; or DAGster?  At the moment, we 'just' want data for ML experimentation.  Early in 2022 we'll also want an 'operational' data pipeline that gets data every 5 minutes :)

peterdudfield:
PV data is current from '2010-12-15T13:10:00. to  '2019-08-20T23:00:00', Ill have a look at updating this

JackKelly:
Awesome, thank you!

That data is from PVOutput.org.

I wrote a little Python library to download data from PVOutput.org: https://github.com/openclimatefix/pvoutput

PVOutput.org provide an API for _some_ of the data we need.  But the script also needs to scrape some data from the PVOutput.org HTML pages.

I haven't used the Python library for something like 2 years so I wouldn't be surprised if the PVOutput.org API and/or website are out of sync.

Also, OCF will need to pay PVOutput.org for a new subscription for a year's access to UK PV data (which costs something like $800 Australian dollars).  There is free API access but it's so slow as to be almost unusable for bulk-downloading lots of data.  I'll pay PVOutput.org this afternoon when I'm back in London and forward you the account details ;)

[...]

PR #172, opened 2021-09-28
Add download script for EUMETSAT Data

jacobbieker:
# Pull Request

## Description

This adds a download script for EUMETSAT RSS and Cloud mask data. Basing it off of SatFlow's initial downloading script. The majority of the code for this has now been moved to https://github.com/openclimatefix/Satip/pull/11, this PR just now contains the Click functionality and wrapper script

Fixes issue #91

Relates to #95 #81 

## How Has This Been Tested?

Running it locally, randomly deleting files and making sure it backfills it.

- [ ] No
- [x] Yes

## Checklist:


JackKelly:
Awesome, thanks loads for working on this!  Very cool!

jacobbieker:
As of now, the script does successfully download, sanity check RSS images, and seems to successfully backfill missing data from any given day. What is missing is sanity checking the cloud masks, as Satpy doesn't seem to want to load just the cloud mask on its own without throwing an error. But overall, it does function for getting data between two dates.

jacobbieker:
There is now a filesize sanity check for cloud masks, which if it fails removes the file

JackKelly:
Sounds like excellent progress!

jacobbieker:
And now backfilling, so I think this script is mostly complete?

jacobbieker:
The main difference between this script and the idea set out in #91 is that `.partial` isn't appended to the file while its being downloaded, the bandwidth limiter doesn't do anything, and its single threaded. As the actual downloading is done in Satip, will open a PR to append `.partial` to files as they download there. This script also downloads chunks of files at once, usually a whole day's worth, before running sanity checks and moving them to their final folders. If the script is interrupted and restarted, it'll run the sanity checks on any files in the download directory before starting up again, so should save progress.

scripts/get_raw_eumetsat_data.py
```diff
@@ -0,0 +1,354 @@
...
+    "-b",
+    default=False,
+    prompt="Whether to download any missing data from the start date of the data on disk to the end date",
+    is_flag=True,
+)
+@click.option(
+    "--user_key",
+    "--key",
+    default=None,
+    help="The User Key for EUMETSAT access",
```
JackKelly, line 79: Maybe add a sentence to the `help` to say something like `Alternatively, the user key can be set using an auth file.` ?

scripts/get_raw_eumetsat_data.py
```diff
@@ -0,0 +1,354 @@
...
+        backfill: Whether to backfill between the beginning of EUMETSAT data and now, overrides start and end date
+        bandwidth_limit: Bandwidth limit, currently unused
+        user_key: User key for the EUMETSAT API
+        user_secret: User secret for the EUMETSAT API
+        auth_filename: Path to a file containing the user_secret and user_key
+
+    """
+    # Get authentication
+    if auth_filename is not None:
+        user_key, user_secret = load_key_secret(auth_filename)
```
JackKelly, line 126: Just to be really careful, it might be good throw an error if `auth_filename` is set _and_ (`user_key` or `user_secret` are already set).  Otherwise the user might be confused why `user_key` and/or `user_secret` are being ignored :)

e.g., something like:

```python
if auth_filename is not None:
    if (user_key is not None) or (user_secret is not None):
        raise RuntimeError("Please do not set BOTH auth_filename AND user_key or user_secret!")
    user_key, user_secret = load_key_secret(auth_filename)
```

JackKelly (commented):
Looks fantastic, thanks so much for doing all this work!

One thought, which I should've had earlier:

Do you think we should move this all into `Satip` because this download script doesn't do anything that's specific to ML, and because there might be a bunch of users out there who would love to be able to download EUMETSAT data, but who have no interest in ML?!?

This review is only half-done!  I'll finish ASAP...

jacobbieker:
> Looks fantastic, thanks so much for doing all this work!
> 
> One thought, which I should've had earlier:
> 
> Do you think we should move this all into `Satip` because this download script doesn't do anything that's specific to ML, and because there might be a bunch of users out there who would love to be able to download EUMETSAT data, but who have no interest in ML?!?
> 
> This review is only half-done! I'll finish ASAP...

Yeah, that would make sense to have it there! I'll open a PR with it, but maybe just leave this one for comments, and I'll incorporate that in the version in Satip? The nice thing of having it here is that we have the other data downloading scripts in nowcasting-dataset as well, so it keeps it in one place. I might change this to a small file with just the click options and call `download_eumetsat_data` from Satip instead

scripts/get_raw_eumetsat_data.py
jacobbieker: Done!

scripts/get_raw_eumetsat_data.py
jacobbieker: Done!

scripts/get_raw_eumetsat_data.py
```diff
@@ -0,0 +1,354 @@
...
+from datetime import datetime, timedelta
+
+from nowcasting_dataset.consts import SAT_VARIABLE_NAMES
+import click
+
+# SatPy gives a UserWarning about Proj, which isn't actually used here, so can be
+# safely ignored
+import warnings
+
+warnings.simplefilter("ignore", UserWarning)
```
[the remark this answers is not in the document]
jacobbieker: Done!

jacobbieker:
> > Looks fantastic, thanks so much for doing all this work!
> > One thought, which I should've had earlier:
> > Do you think we should move this all into `Satip` because this download script doesn't do anything that's specific to ML, and because there might be a bunch of users out there who would love to be able to download EUMETSAT data, but who have no interest in ML?!?
> > This review is only half-done! I'll finish ASAP...
> 
> Yeah, that would make sense to have it there! I'll open a PR with it, but maybe just leave this one for comments, and I'll incorporate that in the version in Satip? The nice thing of having it here is that we have the other data downloading scripts in nowcasting-dataset as well, so it keeps it in one place. I might change this to a small file with just the click options and call `download_eumetsat_data` from Satip instead

The moved script is at https://github.com/openclimatefix/Satip/pull/11 I'll address any more comments you have on it here on that PR

PR #195, opened 2021-10-05
Issue/166 batch pydantic

### LINKED ISSUE #166: Example --> Pydantic
Move [Example](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/dataset/example.py) to Pydantic object

## Detailed Description
The Example data (maybe renamed to 'BatchData') is getting a bit bigger, and lacks a bit of structure. It might be good to move it to something with a bit more structure. If we need move to pydantic, then then we can use the internal validators to validate the data nicely. 

We would have to check that is saves and compresses well

## Possible Implementation
BatchData
 - General
    - t0_dt
    - batch_size
 - PV
   - pv_yield
   - ....
 - {Next_Data_Source}

peterdudfield:
# Pull Request

## Description

- Pydantic model for batch
- Pydantic model for eahc data source
- tools for changing model to xarray (and back)
- adjust DataSets to use pydantic models
 
Fixes #166 

## How Has This Been Tested?

- [ ] No
- [ ] Yes

## Checklist:

### DIFF
--- a/notebooks/2021-10/2021-10-01/pydantic.py
+++ b/notebooks/2021-10/2021-10-01/pydantic.py
@@ -1,0 +1,129 @@
+from pydantic import BaseModel, Field, validator
+from typing import Union
+import numpy as np
+import xarray as xr
+import torch
+from nowcasting_dataset.config.model import Configuration
+
+
+Array = Union[xr.DataArray, np.ndarray, torch.Tensor]
+
+
+class Satellite(BaseModel):
+
+    # width: int = Field(..., g=0, description="The width of the satellite image")
+    # height: int = Field(..., g=0, description="The width of the satellite image")
+    # num_channels: int = Field(..., g=0, description="The width of the satellite image")
+
+    # Shape: [batch_size,] seq_length, width, height, channel
+    image_data: Array = Field(
+        ...,
+        description="Satellites images. Shape: [batch_size,] seq_length, width, height, channel",
+    )
+    x_coords: Array = Field(
+        ...,
+        description="The x (OSGB geo-spatial) coordinates of the satellite images. Shape: [batch_size,] width",
+    )
+    y_coords: Array = Field(
+        ...,
+        description="The y (OSGB geo-spatial) coordinates of the satellite images. Shape: [batch_size,] width",
+    )
+
+    # @validator("sat_data")
+    # def image_shape(cls, v):
+    #     assert v.shape[-1] == cls.num_channels
+    #     assert v.shape[-2] == cls.height
+    #     assert v.shape[-3] == cls.width
+    #
+    # @validator("x_coords")
+    # def x_coords_shape(cls, v):
+    #     assert v.shape[-1] == cls.width
+    #
+    # @validator("y_coords")
+    # def y_coords_shape(cls, v):
+    #     assert v.shape[-1] == cls.height
+    #
+    class Config:
+        arbitrary_types_allowed = True
+
+
+class Batch(BaseModel):
+
+    batch_size: int = Field(
+        ...,
+        g=0,
+        description="The size of this batch. If the batch size is 0, "
+        "then this item stores one data item",
+    )
+
+    satellite: Satellite
+
+
+class FakeDataset(torch.utils.data.Dataset):
+    """Fake dataset."""
+
+    def __init__(self, configuration: Configuration = Configuration(), length: int = 10):
+        """
+        Init
+
+        Args:
+            configuration: configuration object
+            length: length of dataset
+        """
+        self.batch_size = configuration.process.batch_size
+        self.seq_length_5 = (
+            configuration.process.seq_len_5_minutes
+        )  # the sequence data in 5 minute steps
+        self.seq_length_30 = (
+            configuration.process.seq_len_30_minutes
+        )  # the sequence data in 30 minute steps
+        self.satellite_image_size_pixels = configuration.process.satellite_image_size_pixels
+        self.nwp_image_size_pixels = configuration.process.nwp_image_size_pixels
+        self.number_sat_channels = len(configuration.process.sat_channels)
+        self.number_nwp_channels = len(configuration.process.nwp_channels)
+        self.length = length
+
+    def __len__(self):
+        """Number of pieces of data"""
+        return self.length
+
+    def per_worker_init(self, worker_id: int):
+        """Not needed"""
+        pass
+
+    def __getitem__(self, idx):
+        """
+        Get item, use for iter and next method
+
+        Args:
+            idx: batch index
+
+        Returns: Dictionary of random data
+
+        """
+
+        sat = Satellite(
+            image_data=np.random.randn(
+                self.batch_size,
+                self.seq_length_5,
+                self.satellite_image_size_pixels,
+                self.satellite_image_size_pixels,
+                self.number_sat_channels,
+            ),
+            x_coords=torch.sort(torch.randn(self.batch_size, self.satellite_image_size_pixels))[0],
+            y_coords=torch.sort(
+                torch.randn(self.batch_size, self.satellite_image_size_pixels), descending=True
+            )[0],
+        )
+
+        # Note need to return as nested dict
+        return Batch(satellite=sat, batch_size=self.batch_size).dict()
+
+
+train = torch.utils.data.DataLoader(FakeDataset())
+i = iter(train)
+x = next(i)
+
+x = Batch(**x)
+# IT WORKS
+assert type(x.satellite.image_data) == torch.Tensor
--- a/nowcasting_dataset/config/model.py
+++ b/nowcasting_dataset/config/model.py
@@ -107,7 +107,7 @@ class Process(BaseModel):
     )
     history_minutes: int = Field(30, ge=0, description="how many historic minutes are used")
     satellite_image_size_pixels: int = Field(64, description="the size of the satellite images")
-    nwp_image_size_pixels: int = Field(2, description="the size of the nwp images")
+    nwp_image_size_pixels: int = Field(64, description="the size of the nwp images")
 
     sat_channels: tuple = Field(
         SAT_VARIABLE_NAMES, description="the satellite channels that are used"
--- a/nowcasting_dataset/consts.py
+++ b/nowcasting_dataset/consts.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Union
 
 import numpy as np
+import torch
 import xarray as xr
 
 # DEFAULT PATHS
@@ -23,7 +24,7 @@ PV_METADATA_FILENAME = PV_PATH / "UK_PV_metadata.csv"
 NWP_FILENAME = "gs://" + str(BUCKET / "NWP/UK_Met_Office/UKV_zarr")
 
 # Typing
-Array = Union[xr.DataArray, np.ndarray]
+Array = Union[xr.DataArray, np.ndarray, torch.Tensor]
 PV_SYSTEM_ID: str = "pv_system_id"
 PV_SYSTEM_ROW_NUMBER = "pv_system_row_number"
 PV_SYSTEM_X_COORDS = "pv_system_x_coords"
--- a/nowcasting_dataset/dataset/model/datasource.py
+++ b/nowcasting_dataset/dataset/model/datasource.py
@@ -1,19 +1,61 @@
-from pydantic import BaseModel
+import torch
+from pydantic import BaseModel, Field
 import pandas as pd
 import xarray as xr
 import numpy as np
 
+from typing import List
 
-class DataSource(BaseModel):
+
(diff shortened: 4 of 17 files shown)


nowcasting_dataset/data_sources/datetime/datetime_model.py
```diff
@@ -0,0 +1,100 @@
...
+
+    hour_of_day_sin: Array  #: Shape: [batch_size,] seq_length
+    hour_of_day_cos: Array
+    day_of_year_sin: Array
+    day_of_year_cos: Array
+    datetime_index: Array
+
+    @property
+    def sequence_length(self):
+        """The equence length of the pv data"""
```
jacobbieker, line 22: tiny typo: sequence

nowcasting_dataset/data_sources/gsp/gsp_model.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+    @property
+    def number_of_gsp(self):
+        """The number of pv systems"""
+        return self.gsp_yield.shape[-1]
+
+    @property
+    def sequence_length(self):
+        """The equence length of the pv data"""
```
jacobbieker, line 59: tiny typo: sequence

nowcasting_dataset/data_sources/nwp/nwp_model.py
```diff
@@ -0,0 +1,161 @@
...
+        return self.nwp.shape[-2]
+
+    @property
+    def height(self):
+        """The width of the nwp data"""
+        return self.nwp.shape[-1]
+
+    @property
+    def sequence_length(self):
+        """The equence length of the pv data"""
```
jacobbieker, line 66: sequence

nowcasting_dataset/dataset/batch.py
```diff
@@ -3,209 +3,273 @@
...
-                        **{"pv_system": np.arange(n_pv_systems, dtype=np.int32)},
-                    },
-                    dims=["example", "pv_system"],
-                )
+    # loop over each item in the batch
+    for i, example in enumerate(batch.split()):
+
+        individual_datasets = []
+
+        individual_datasets.append(example.satellite.to_xr_dataset())
```
jacobbieker, line 246: Is there a more concise way we could loop through all the different datasets?

nowcasting_dataset/dataset/datasets.py
```diff
@@ -429,12 +408,8 @@ def subselect_data(
     )
 
     # t0_dt or if not available use a different datetime index
-    if T0_DT in batch.keys():
-        current_time_of_first_batch = batch[T0_DT][0]
-    else:
-        current_time_of_first_batch = batch[date_time_index_to_use].isel(
-            time=current_timestep_index
-        )[0]
+    batch.general.to_numpy()
```
jacobbieker, line 411: Is there a way we could keep it as xarrays? changing it back and forth from NumPy adds some overhead to it.

jacobbieker (commented):
Looking good! Just a few typos and minor things, but I like how its done!

nowcasting_dataset/data_sources/README.md
```diff
@@ -0,0 +1,37 @@
+This folder contains the code for the different data sources.
+
+# Data Sources
+- datetime: datetime information like 'hour_of_day'
+- general: metadata for the batch like t0_dt
+- gsp: Grid Supply Point data from Sheffield Solar
```
JackKelly, line 6:
```suggestion
- gsp: Grid Supply Point data from Sheffield Solar (e.g. the estimated total solar PV power generation for each GSP region, and the geospatial shape of each GSP region).
```

nowcasting_dataset/data_sources/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+# datasource_output.py
+
+General pydantic model of output of the data source. Contains the following methods
+- to_numpy: changes all data points to numpy objects
+- split: converts a batch to a list of items
+- join: joins list of items to one
+- to_xr_dataset: changes data items to xarrays and returns a dataset
+- from_xr_dataset: loads from a xr dataset
```
JackKelly, line 29:
```suggestion
- from_xr_dataset: loads from an xarray dataset
```

(just in case people don't know what `xr` stands for!)

nowcasting_dataset/data_sources/README.md
```diff
@@ -0,0 +1,37 @@
...
+- select_time_period: subselect data, depending on a time period
+
+# <X> Data Source folder
+
+Roughly each of the data source folders follows this pattern
+- A class which is how to load the data source, how to select for batches e.t.c. This is built from 'data_source.py',
```
JackKelly, line 35:
```suggestion
- A class which defines how to load the data source, how to select for batches etc. This inherits from 'data_source.DataSource',
```

nowcasting_dataset/data_sources/README.md
```diff
@@ -0,0 +1,37 @@
...
+- A class which contains the output model of the data source. This is the information used in the batches.
+This is built from 'datasource_output.py',
```
JackKelly, line 37:
```suggestion
This inherits from 'datasource_output.DataSourceOutput'.
```

nowcasting_dataset/data_sources/data_source.py
```diff
@@ -99,7 +99,7 @@ def get_batch(
         t0_datetimes: pd.DatetimeIndex,
         x_locations: Iterable[Number],
         y_locations: Iterable[Number],
-    ) -> List[Example]:
+    ) -> DataSourceOutput:
```
JackKelly, line 102: Nice!

nowcasting_dataset/data_sources/data_source.py
```diff
@@ -112,15 +112,16 @@ def get_batch(
         Returns: Batch data
 
         """
-        examples = []
+        data = []
```
JackKelly, line 115: Maybe I'm just being resistent to change(!) :) But, I must admit that, in this function, I feel it would be a tiny bit easier to follow if we renamed `data` to `examples`; and renamed `output` to `example`.  My thinking is that, in the context of this function, an `output` will always be a single training example; and `data` will always be a list of examples.

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+
+from nowcasting_dataset.utils import to_numpy
+
+logger = logging.getLogger(__name__)
+
+
+class DataSourceOutput(BaseModel):
+    """General Data Source output pydantic class.
+
+    Data source output classes should build of this class
```
JackKelly, line 17:
```suggestion
    Data source output classes should inherit from this class
```

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    class Config:
+        """ Allowed classed i.e. tensor.Tensor"""
```
JackKelly, line 21:
```suggestion
        """ Allowed classed e.g. torch.Tensor"""
```

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+    batch_size: int = Field(
+        0,
+        ge=0,
+        description="The size of this batch. If the batch size is 0, "
+        "then this item stores one data item",
+    )
+
+    def to_numpy(self):
+        """Change to numpy"""
+        for k, v in self.dict().items():
```
JackKelly, line 35: nice!

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+            self.__setattr__(k, to_numpy(v))
+
+    def to_xr_data_array(self):
+        """ Change to xr DataArray"""
+        pass
```
JackKelly, line 40: Assuming we want sub-classes to define this function, then I'd do:

```suggestion
        raise NotImplementedError()
```

Unless we really do want `to_xr_data_array()` to silently do nothing??

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @staticmethod
+    def join(data):
```
JackKelly, line 43: What's the expected data type of `data`?

And, instead of the function name `join`, maybe we could be even more explicit about this function's intended purpose and call it something like `create_batch_from_examples()`?

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+        """ Join a list of data source items to a batch. """
+        _ = [d.to_numpy() for d in data]
+
+        # use the first item in the list, and then update each item
+        batch = data[0]
+        for k, v in batch.dict().items():
```
JackKelly, line 49: I don't think `v` is used, so this could become:

```suggestion
        for k in batch.dict():
```

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return batch
+
+    def split(self):
+        """
+        Split the datasource from a batch to a list of items
+
+        Returns: List of single data source items
+        """
+        c = self.__class__
```
JackKelly, line 68: I'd be tempted to rename `c` to `cls` or something like that?  I almost missed the `c` on line 74!

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def to_xr_dataset(self):
+        """ Make a xr dataset. Each data source needs to defined this """
+        raise NotImplementedError
+
+    def from_xr_dataset(self):
+        """ Load from xr dataset. Each data source needs to defined this """
+        raise NotImplementedError
+
+    def select_time_period(
```
JackKelly, line 86: So, I could be wrong, but I think the main use-case for this sort of function is to limit the number of timesteps of each example.  For example, perhaps the pre-prepared batches on disk have a history length of 2 hours, and a forecast length of 4 hours.  But a particular model only wants an hour of history and an hour of forecast.

So, instead of asking the user to specify a `time_of_first_example`, `start_time`, and `end_time`; it may instead be more useful if the user just provides a new `history_minutes` and `forecast_minutes`, and the returned batch is cropped.

Our imaginary user who only wants an hour of history and an hour of forecast would just call `select_time_period(history_miniutes=60, forecast_minutes=60)` and they'd get back a cropped batch.

But, perhaps there's another use-case that I've forgotten about?!

JackKelly (commented):
OOPS, SORRY!  I SENT TOO EARLY!  LET ME EDIT!


In general, and perhaps I'm being too resistent to change :) but I think I prefer the variable name `example` over `data item`.  I think `data item` is less descriptive and more vague than `example`!  But I do agree that `example` is perhaps a bit of a wierd name.  Maybe `training_example` is better?  (Although each example can also be used for testing and validation; not just _training_!)

One "safety net" that was provided by forcing all the modalities into an `xr.Dataset` was that it kind of guaranteed that every modality was aligned in time and space.  Is there a way to ensure that each modality is aligned in time and space in the new framework?

nowcasting_dataset/data_sources/data_source.py
peterdudfield: thanks

nowcasting_dataset/data_sources/datasource_output.py
jacobbieker: I think that makes sense, the only other use case I can think of is if the imaginary user wants to change the current time of the batch, and take the current time as both an hour in the past, as well as 2 hours in the future as well, to get more examples out of each batch.

nowcasting_dataset/data_sources/data_source.py
peterdudfield: I'm happy to keep it as example

nowcasting_dataset/data_sources/datasource_output.py
```diff
@@ -0,0 +1,155 @@
...
+                batch.batch_size = len(data)
+            else:
+
+                # get list of one variable from the list of data items.
+                one_variable_list = [d.__getattribute__(k) for d in data]
+                batch.__setattr__(k, np.stack(one_variable_list, axis=0))
+
+        return batch
+
+    def split(self):
```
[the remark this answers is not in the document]
peterdudfield: yea, can do something like, but its not stricly truth as the inhertited class will not return DataSourceOutput it will return, for example 'GSP'. Maybe there is no way round this

nowcasting_dataset/dataset/datasets.py
peterdudfield: Ive changed it so it just takes .value to t0_dt

PR #229, opened 2021-10-14
Issue/209 xrdatarray b

### LINKED ISSUE #209: Can we simplify the code by always keeping the data in one data type (e.g. xr.DataArray) per modality?
In the past, `nowcasting_dataset` was designed to feed data on-the-fly into a PyTorch model.  Which meant that, as the data flowed through `nowcasting_dataset`, the data would change type:  For example, satellite data would start as an `xr.DataArray`, then get turned into a `numpy array` (because PyTorch doesn't know what to do with an `xr.DataArray`), and then get turned into a `torch.Tensor`.

But, I think we can safely say now that `nowcasting_dataset` is _just_ for pre-preparing batches (_not_ for loading data on-the-fly).  As such, we can probably simplify the code by keeping data in a single container type per modality.  For example, satellite data could _always_ live in an `xr.DataArray` for its entire life while flowing through `nowcasting_dataset`.

Sorry, I really should've thought of this earlier!  But, yeah, I think this could simplify the code quite a lot.

I haven't fully thought through the implications of this, but some changes might be:

* In the Pydantic models, each field can be just a _single_ type (instead of a `Union` of types).  So, for example, instead of `sat_data: Array = Field(...` we can just do `sat_data: xr.DataArray = Field(...`
* We can get rid of the `to_numpy` function.
* For all the modalities which use xarray of pandas data types, we can use dimension _names_ instead of _indexes_.  e.g. `seq_length = len(sat_data[-4])` becomes `seq_length = len(sat_data.time)`
* Saving and load data to/from disk becomes super-simple.
* We'd no longer need the `to_xr_dataset` and `from_xr_dataset` methods (which are quite fiddly)

peterdudfield:
# Pull Request

## Description

There are two different flows for the data representting, 1. saving to a batch file, 2. loading ready for ml

1. 
 - Each data source outputs a xr.Dataset which has been validated using a pydantic extension
 - Batches are made and then saved to .netcdf, each datasource in a different file

2. 
- Batches are loaded from netcdf
- transfered to BatchML, which is pyadntic model of each data source, and each data source describes which tensor fields it should have

*** note that 2. could move to dataloader repo, but wanted to get it all sorted here first

Screenshot 2021-10-15 at 10 56 25

- [x] Change batch maker to use xr.Datasets
- [x] validate all data source xr.Datasets to have the correct dims
- [x] add validation to xr.Datasets for each data source (done for satellite and nwp)
- [x] Create BatchML ready for ML training e.t.c

Fixes #209

## How Has This Been Tested?

- Adjust unittest accordingly
- written new unittest for new code
- ran script/prepare_ml_data.py - it works!!!

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/nwp/nwp_data_source.py
+++ b/nowcasting_dataset/data_sources/nwp/nwp_data_source.py
@@ -233,6 +233,9 @@ class NWPDataSource(ZarrDataSource):
         selected_data = selected_data.sel(target_time=slice(start_dt, end_dt))
         selected_data = selected_data.rename({"target_time": "time"})
         selected_data = selected_data.rename({"variable": "channels"})
+
+        selected_data.data = selected_data.data.astype(np.float32)
+
         return selected_data
 
     def datetime_index(self) -> pd.DatetimeIndex:
--- a/nowcasting_dataset/data_sources/satellite/satellite_data_source.py
+++ b/nowcasting_dataset/data_sources/satellite/satellite_data_source.py
@@ -175,6 +175,9 @@ class SatelliteDataSource(ZarrDataSource):
         if self.normalise:
             selected_data = selected_data - SAT_MEAN
             selected_data = selected_data / SAT_STD
+
+        selected_data.data = selected_data.data.astype(np.float32)
+
         return selected_data
 
     def datetime_index(self, remove_night: bool = True) -> pd.DatetimeIndex:


peterdudfield:
> wow, what a beast!
> 
> I think this PR is waaaaay too big. We should see if we can split up future work better to end up with smaller PRs. I find it really hard to review PRs this size and it takes quite a long time.
> 
> Google's eng-practices make a good case for smaller PRs (they call them CL: Change list): https://google.github.io/eng-practices/review/developer/small-cls.html
> 
> Also, my github browser window got really slow during reviewing haha.
> 
> I agree with @jacobbieker that we should try to remove commented-out code, as it's contained in the git history anyways.
> 
> There are also a bunch of unresolved todo items that are being introduced - I think we should generally try to either:
> 
> a) resolve those before merging the PR, or b) raise an issue or link them to one, and then reference the issue number in the todo comment.
> 
> Otherwise I feel like we'll just forget about them or it's unclear who will work on resolving them in the future.

Yea I agree, this was a big one. I really dont like doing big PRs, so I totally agree about breaking them down. 

Thanks very much @flowirtz for all the comments

PR #329, opened 2021-11-03
Issue/86 remove torch

### LINKED ISSUE #86: Remove PyTorch from the code
When I started `nowcasting_dataset`, the intention was to use `nowcasting_dataset` to generate batches on-the-fly during ML training from separate Zarr stores for the satellite data, NWPs, and PV.  But that turned out to be too slow and fragile :)  So, we swapped to using `nowcasting_dataset` to pre-prepare batches ahead-of-time, and save them to disk.   During ML training, we just need to load the batches from disk, and we're good-to-go.  (Pre-preparing batches has a number of other advantages, too).

But, this development history means that `nowcasting_dataset` still uses PyTorch (e.g. using the PyTorch DataLoader to run multiple processes).  The code may become cleaner and faster and more flexible if we strip out PyTorch, and instead (maybe) use [`concurrent.futures.ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) to use multiple processes.

TODO:

- [x] Remove Datasets and Datamodule.  Done in PR #307 
- [x] Remove pytorch lightning from requirements.txt and environment.yaml.  Done in PR #307.
- [ ] Remove torch from other python files
- [ ] Remove torch from requirements.txt and environment.yaml

peterdudfield:
# Pull Request

## Description

Remove torch code from package

Have not removed torch code from old notebooks

Fixes #86 

## How Has This Been Tested?

Search the folders 
- nowcasting_dataset
- tests

for the word 'torch' - its not there

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,6 @@ conda activate nowcasting_dataset
 pip install -e .
 ```
 
-Note you can install the [pytorch](https://github.com/pytorch/pytorch)
-and [pytorch_lightning](https://github.com/PyTorchLightning/pytorch-lightning) using
-```shell
-pip install -e .[torch]
-```
-but it is only used to create a dataloader for machine learning models, and will not be necessary
-soon (when the dataloader is moved to `nowcasting_dataloader`).
-
-
 ### `pip`
 
 A (probably older) version is also available through `pip install nowcasting-dataset`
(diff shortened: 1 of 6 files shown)


jacobbieker (approved):
LGTM!

PR #332, opened 2021-11-03
Remove Datetime Data Source

### LINKED ISSUE #208: Compute datetime features on-the-fly.  Remove them from the on-disk batches.
It's possible that we might want to remove datetime features from the on-disk batches, and instead compute datetime features on-the-fly, not least because we'll want to experiment with a variety of different ways of encoding position when we really start experimenting with the perceiver IO models (see openclimatefix/nowcasting_utils#30)

These datetime features should be really fast to compute.  And, if we start saving different 'modalities' to disk as different files, then it will almost certainly be faster to compute these on-the-fly rather than saving them into tiny files on disk.

jacobbieker:
# Pull Request

## Description

Removes the Datetime Data Source, as that isn't used anymore and the features are now computed on the fly in the dataloader.

Fixes #208 

## How Has This Been Tested?

Unit tests

- [ ] No
- [x] Yes

## Checklist:


JackKelly:
Nice!

I think there are also some other functions which compute datetime features and can be removed (along with any tests):

* utils.sin_and_cos
* time.datetime_features
* time.datetime_features_in_example

peterdudfield (commented):
Looks good, Thanks for doing this

Is there any readme's to update?

Would it possible to get a one liek this ready for dataloader at the same time, then we can merge at the same point, and hopefully not have any problems

jacobbieker:
> Looks good, Thanks for doing this
> 
> Is there any readme's to update?
> 
> Would it possible to get a one liek this ready for dataloader at the same time, then we can merge at the same point, and hopefully not have any problems

Yeah! I'll start on that

jacobbieker:
Its here: https://github.com/openclimatefix/nowcasting_dataloader/pull/47

PR #334, opened 2021-11-03
adjust save method to deal with Pathy objects, and add method 'manage…

### LINKED ISSUE #316: Copy config YAML to `<target directory>/metadata` and include `git` info in separate file
## Detailed Description
As discussed in this morning's call:

Separate out the `git` info from the `Configuration` object so we can keep a record of the `Config` just by copying the file to the target directory; and put the `git` data into a separate file.  All in a `<target dir>/metadata` directory.

This means we can use rich objects in the `Config` object.  And we can remove the `save_yaml_config` functionality and tests.

Let's do this as a separate PR instead of implementing this in my massive PR :slightly_smiling_face:

peterdudfield:
…r.py'

# Pull Request

## Description

- Save config to output data file
- found method to save pathy object easily

Fixes #316 

## How Has This Been Tested?

- normall unittests
- added a unittest back in

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/config/__init__.py
+++ b/nowcasting_dataset/config/__init__.py
@@ -1,3 +1,4 @@
 """ Configuration of the dataset """
 from nowcasting_dataset.config.load import load_yaml_configuration
 from nowcasting_dataset.config.model import Configuration, InputData, set_git_commit
+from nowcasting_dataset.config.save import save_yaml_configuration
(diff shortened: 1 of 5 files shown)


JackKelly (approved):
That's a very elegant solution, thanks!  LGTM!

PR #335, opened 2021-11-04
nwp float16

### LINKED ISSUE #61: Use smaller dtypes for saved data
int16 for both NWP & satellite?  Maybe even uint8 (and use 0 for NaN???)

peterdudfield:
# Pull Request

## Description

- Chang nwp to float16
- some fixes to the test data generation scripts
- update nwp test data

Fixes #61 

## How Has This Been Tested?

Normal unitests, 

nwp test data went from ~ 13M --> 5M

- [ ] No
- [x] Yes

## Checklist:


JackKelly (approved):
The satellite bit looks good!

But the NWP bit is a lot more involved so I'd leave NWPs as float32 for now. Before converting to int 16, I the codes need to rescale to 0 to 2**15 . Although, when we do that, we can probably go  to 8 bit for nwp

peterdudfield:
> The satellite bit looks good!
> 
> But the NWP bit is a lot more involved so I'd leave NWPs as float32 for now. Before converting to int 16, I the codes need to rescale to 0 to 2**15 . Although, when we do that, we can probably go to 8 bit for nwp

If I remove the nwp conversion, then I don't think there is any point merging. The satellite doesn't reduce in size. 

Do you mind expanding on what needs doing for the NWP data?

JackKelly:
> Do you mind expanding on what needs doing for the NWP data?

Sure!

The short answer is that we want to maintain as much information as possible.

I'll illustrate with a slightly cartoonish worst-case example:

Let's pretend that NWPs measured irradiance in units of kW per meter squared.  If that were true, then the values for irradiance would always be between 0.0 and about 0.7 kW/m<sup>2</sup>.

If we just did `irradiance.astype(np.int16)` then every value would be converted to 0.  We'd lose _all_ the information!

Instead, we want to "stretch" the irradiance values so they most best use of int16.  int16 can represent values in the range [-32,768, 32,767].  But, for ease, we probably just want to use the positive numbers [0, 32,767]; and use -1 to represent NaNs.

So, we need to do something like this:

```python
# Rescale to [0, 1]:
nwp -= nwp.min()
nwp /= nwp.max()

# Rescale to [0, 32,767]:
nwp *= 2**15 - 1

# Fill NaNs with -1 (because int16 has no concept of "NaN")
nwp = nwp.fillna(-1)

# Convert to int16
nwp = nwp.astype(np.float16)
```

But, in order to rescale like this, we need to know the min and max for every NWP variable, across every value in time and space.

Then, in `nowcasting_dataloader`, we'd need to normalise using the standard deviation and mean computed using the rescaled values.

peterdudfield:
> > Do you mind expanding on what needs doing for the NWP data?
> 
> Sure!
> 
> The short answer is that we want to maintain as much information as possible.
> 
> I'll illustrate with a slightly cartoonish worst-case example:
> 
> Let's pretend that NWPs measured irradiance in units of kW per meter squared. If that were true, then the values for irradiance would always be between 0.0 and about 0.7 kW/m2 in the UK.
> 
> If we just did `irradiance.astype(np.int16)` then every value would be converted to 0. We'd lose _all_ the information!
> 
> Instead, we want to "stretch" the irradiance values so they fill int16's full range. int16 can represent values in the range [-32,768, 32,767]. But, for ease, we probably just want to use the positive numbers [0, 32,767]; and use -1 to represent NaNs.
> 
> So, we need to do something like this:
> 
> ```python
> # Rescale to [0, 1]:
> nwp -= nwp.min()
> nwp /= nwp.max()
> 
> # Rescale to [0, 32,767]:
> nwp *= 2**15 - 1
> 
> # Fill NaNs with -1 (because int16 has no concept of "NaN")
> nwp = nwp.fillna(-1)
> 
> # Convert to int16
> nwp = nwp.astype(np.float16)
> ```
> 
> But, in order to rescale like this, we need to know the min and max for every NWP variable, across every value in time and space.
> 
> Then, in `nowcasting_dataloader`, we'd need to normalise using the standard deviation and mean computed using the rescaled values.

Ok, I think ill do the satellite int16 and then think again use the nwp data a the way you have described

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
```diff
@@ -55,6 +56,8 @@ def _get_time_slice(self, t0_dt: pd.Timestamp) -> xr.DataArray:
         start_dt = self._get_start_dt(t0_dt)
         end_dt = self._get_end_dt(t0_dt)
         data = self.data.sel(time=slice(start_dt, end_dt))
+        data = data.astype(np.int16)
```
JackKelly, line 59: What would you think about _not_ including this line (and hence `nowcasting_dataset` would just propagate the satellite data type from the incoming Zarr)?

On the one hand, keeping this here makes it more explicit.

On the other hand, you could imagine a situation where someone updates `Satip` to use a different dtype for the satellite Zarr (e.g. int8 or float16) and we might forget to update the dtype here, which might result in weird bugs.   It might be nice to only have to change the dtype in one place in the code, and to have clearly defined responsibilities between the different repos?  (e.g. we could declare that `nowcasting_dataset` never changes the numerical values: it just chops up the data)

JackKelly (approved):

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
peterdudfield: probably best to remove it for now - so we just default what is in `Satip`

JackKelly: SGTM!

Issue #319, opened 2021-11-02
Standardise naming for `t0_datetime_UTC`, `x_center_OSGB`, `y_center_OSGB`


JackKelly:
* `DataSource.create_batches()` uses `t0_datetime_UTC`, `x_center_OSGB`, `y_center_OSGB` 
* `DataSource.get_batch()` uses `t0_datetimes`, `x_locations`, `y_locations`
* `DataSource.get_example()` uses `t0_dt`, `x_meters_center`, `y_meters_center`
* Other `DataSource` methods use other parameter names.

Also update docs in `data_sources/README.md`

I'm not entirely sure what's best but perhaps I'm leaning towards being verbose and using: `t0_datetime_UTC`, `x_center_OSGB`, `y_center_OSGB`, or something like that???  (Although do we _want_ different parameter names for singular objects (the single datetime fed into `get_example()` and plural (the `pd.DatetimeIndex` fed into `get_batch()`)

(This is my fault - I think these parameter names are from when I originally wrote `DataSource` months ago!)

peterdudfield:
should it be 
- t0_datetime_utc
- x_center_osgb
- y_center_osgb?

jacobbieker:
That sounds good to me

JackKelly:
SGTM!

Issue #48, opened 2021-07-07
refactor DataSources to have different functions for selecting timesteps, selecting geo locations, and post-processing


JackKelly:
And reduce duplication between NWPDataSource and SatelliteDataSource, because they both work in very similar ways now!

PR #339, opened 2021-11-05
Update SateliteDataSource to new Zarr format

### LINKED ISSUE #313: Update Satellite Zarr reader for new format
## Detailed Description
Satellite data is now stored in the zarr in unprojected format, so the original geo projection. This does make it slightly more difficult to save the OSGB coordinates in the zarr, as each pixel has a unique OSGB x and y coordinate. Currently, they are stored in the metadata field, under 'osgb_x' and 'osgb_y'. There might be a better way to store it, which we could also do, but for now, we probably need to read in the metadata of one example each for the HRV channel and all the other channels, to get the mapping from image coordinates to OSGB coordinates.

jacobbieker:
# Pull Request

## Description

Zarr files are now split into HRV and non-HRV files, and have their timestamps already rounded to the nearest 5 minutes. This PR just updates the data loading to use these new files.

While there is a lot of changes, most of it is updating the test data to be on 2020-04 instead of 2019-01, as the only satellite data we have in the new form is from 2020-01 onwards

Fixes #313 

## How Has This Been Tested?

Unit tests

- [ ] No
- [x] Yes

## Checklist:


jacobbieker:
Still need to test this on the prepare_ml_script, but is pretty much ready to go otherwise!

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
```diff
@@ -56,11 +59,87 @@ def _get_time_slice(self, t0_dt: pd.Timestamp) -> xr.DataArray:
...
+
+        Ignores x and y coordinates as for the original satellite projection each pixel varies in
+        both its x and y distance from other pixels. See Issue 401 for more details.
+
+        This results, in 'real' spatial terms, each image covering about 2x as much distance in the
+        x direction as in the y direction.
+
+        Args:
+            data_array: DataArray to subselect from
+            x_meters_center: Center of the image x coordinate in OSGB coordinates
```
peterdudfield, line 80: shall we trying to do https://github.com/openclimatefix/nowcasting_dataset/issues/319 when possible
i.e x_center_osgb

nowcasting_dataset/data_sources/satellite/satellite_model.py
```diff
@@ -26,3 +26,11 @@ def model_validation(cls, v):
         )
 
         return v
+
+
+class HRVSatellite(Satellite):
```
peterdudfield, line 31: nice

scripts/generate_data_for_tests/generate_satellite_test_data.py
```diff
@@ -9,21 +10,46 @@
...
+
+# Non-HRV path
+SAT_FILENAME = os.path.join(
+    "/mnt/storage_ssd_8tb/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/satellite/EUMETSAT/SEVIRI_RSS/zarr/v2/eumetsat_zarr_*"
+)
+
 
 def generate_satellite_test_data():
     """Main function to make satelllite test data"""
+    output_filename = OUTPUT_PATH / "hrv_sat_data.zarr"
```
peterdudfield, line 31: do you mind adding a few comments in here, just make it supper clear whats going on

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
jacobbieker: Yeah, I can change it to that

peterdudfield (commented):
Looks all very good, thanks for all this work - 

One thing that sproabbly worth doing is adding it to one of the 'test_manager' and checking a batch file is produced. I think you might have to add the new data sources to the list of data sources [here](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/data_sources/__init__.py)

jacobbieker:
> Looks all very good, thanks for all this work -
> 
> One thing that sproabbly worth doing is adding it to one of the 'test_manager' and checking a batch file is produced. I think you might have to add the new data sources to the list of data sources [here](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/data_sources/__init__.py)

Done!

jacobbieker:
Okay, now all the test data is updated, and all satellite tests pass, only issue is with the NWP test data, of converting to int32 from int64. But otherwise, should be good to go!

jacobbieker:
> Okay, now all the test data is updated, and all satellite tests pass, only issue is with the NWP test data, of converting to int32 from int64. But otherwise, should be good to go!

And its fixed locally! All tests should be passing

jacobbieker:
Only issue that has come up is for Feb 2020, its a leap year and Xarray says that it cannot keep the time dim in numpy datetime64[ns], so converts it to CFTime. This causes a bit of an issue when loading it with all the other Zarrs. Without Feb 2020 it all works fine opening all the other zarrs together and using them.

jacobbieker:
> Only issue that has come up is for Feb 2020, its a leap year and Xarray says that it cannot keep the time dim in numpy datetime64[ns], so converts it to CFTime. This causes a bit of an issue when loading it with all the other Zarrs. Without Feb 2020 it all works fine opening all the other zarrs together and using them.

Gives an error of: 
```
/home/jacob/miniconda3/envs/ocf/lib/python3.9/site-packages/xarray/core/indexing.py:419: SerializationWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using cftime.datetime objects instead, reason: dates out of range
  return np.asarray(array[self.key], dtype=None)
```

nowcasting_dataset/config/model.py
```diff
@@ -115,10 +115,34 @@ class Satellite(DataSourceMixin):
...
-    satellite_image_size_pixels: int = IMAGE_SIZE_PIXELS_FIELD
-    satellite_meters_per_pixel: int = METERS_PER_PIXEL_FIELD
+    satellite_image_size_pixels: int = Field(
+        IMAGE_SIZE_PIXELS_FIELD.default // 3,
+        description="The number of pixels of the region of interest for non-HRV satellite "
+        "channels.",
+    )
+    satellite_meters_per_pixel: int = Field(
+        METERS_PER_PIXEL_FIELD.default * 3,
+        description="The number of meters per pixel for " "non-HRV satellite channels.",
```
JackKelly, line 127: Tiny nit (almost certainly introduced by Black!)

```suggestion
        description="The number of meters per pixel for non-HRV satellite channels.",
```

nowcasting_dataset/config/model.py
```diff
@@ -327,6 +353,7 @@ def set_base_path(self, base_path: str):
             "pv.pv_filename",
             "pv.pv_metadata_filename",
             "satellite.satellite_zarr_path",
+            "satellite_hrv.satellite_zarr_path",
```
JackKelly, line 356: I may be getting confused, but should this be:

```suggestion
            "hrvsatellite.hrvsatellite_zarr_path",
```

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
```diff
@@ -56,11 +59,87 @@ def _get_time_slice(self, t0_dt: pd.Timestamp) -> xr.DataArray:
         start_dt = self._get_start_dt(t0_dt)
         end_dt = self._get_end_dt(t0_dt)
         data = self.data.sel(time=slice(start_dt, end_dt))
-
         assert type(data) == xr.DataArray
 
         return data
 
+    def get_image_pixel(
```
JackKelly, line 66: I'd suggest changing the name of this function... `get_image_pixel()` sounds like it literally gets a single pixel!

Maybe something like `get_square_crop_of_image()`?  Although maybe that's a big clumsy and verbose! 

Maybe `get_spatial_region_of_interest()`?  (I think I slightly prefer that to `get_square_crop_of_image()` because it's more general...)

And, in general, maybe _every_ `DataSource` should have separate steps to `get_time_slice()` and `get_spatial_region_of_interest()`?  (Which has been on my mind for a while:  See #48!)

jacobbieker:
Dropping `acq_time` seems to have fixed it, so might just need to delete that from the Feb Zarr

JackKelly (approved):
LGTM!  Awesome work getting this all working again!  Very exciting!

nowcasting_dataset/config/model.py
jacobbieker: Yeah, you are right! I got a bit confused

nowcasting_dataset/data_sources/satellite/satellite_data_source.py
jacobbieker: Done! And yeah. splitting it out I think is a good idea, but maybe leave it to next year?

JackKelly: > splitting it out I think is a good idea, but maybe leave it to next year?

definitely :slightly_smiling_face:

PR #342, opened 2021-11-08
Issue/340 passive

### LINKED ISSUE #340: Format Passive Data
## Detailed Description
Format Passive data into a similar structure to PV data,
See email train for where the data is stored

## Context
Useful data for the models to help train off

peterdudfield:
# Pull Request

## Description

Format passive data so it can be used as PV

Passive data has been placed
- /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/passive.netcdf
- and on gcs
- updated sun data too, as need to use passive coords

Fixes #340 

## How Has This Been Tested?

- usual unittests
- ran testing [script](https://github.com/openclimatefix/nowcasting_dataset/blob/issue/340-passive/scripts/passiv/passive.py)
- TODO run prepare_ml_data.py - test it on gcp - not run due to

- [ ] No
- [x] Yes

## Checklist:


scripts/passiv/passiv_data.py
```diff
@@ -0,0 +1,77 @@
...
+    passive_5min_df["datetime"] = pd.to_datetime(passive_5min_df["timestamp"])
+
+    # pivot on ssid
+    passive_5min_on_ssid = passive_5min_df.pivot(
+        index="datetime", columns="ss_id", values="power_w"
+    )
+    passive_5min_on_ssid.columns = [str(col) for col in passive_5min_on_ssid.columns]
+
+    # change to xarray
+    passive_5min_jan_xr = passive_5min_on_ssid.to_xarray()
```
peterdudfield, line 52: remove jan

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -368,7 +368,14 @@ def load_solar_pv_data(
     # Convert the pv_system_id column names from strings to ints:
     pv_power_df.columns = [np.int32(col) for col in pv_power_df.columns]
 
-    return pv_power_df.tz_localize("Europe/London").tz_convert("UTC").tz_convert(None)
+    if "passiv" in filename:
+        pv_power_df = pv_power_df.tz_localize("UTC").tz_convert(None)
```
JackKelly, line 372: I might have misunderstood but does this line actually change anything :slightly_smiling_face: ?

Maybe it's sufficient to simplify this `if` block to something like:

```python
if "passive" not in filename:
    # Change PVOutput.org data to UTC and then convert to tz-naive.
    pv_power_df = pv_power_df.tz_localize("Europe/London").tz_convert("UTC").tz_convert(None)
```

scripts/passiv/passiv_data.py
```diff
@@ -0,0 +1,77 @@
...
+    data_df = []
+    for year in years:
+        for month in months:
+            file = f"{dir}/{year}/{month}.csv.gz"
+            print(file)
+            try:
+                df = pd.read_csv(file)
+                data_df.append(df)
+            except Exception as e:  # noqa F481
+                print(f"Could not load {file}")
```
JackKelly, line 32: Do we also want to re-raise this exception, so the script crashes if we can't load this file?

If not, please add a comment to explain why we don't want to crash.

scripts/passiv/passiv_metadata.py
```diff
@@ -0,0 +1,47 @@
...
+
+passive_metadata.to_csv(f"{output_dir}/system_metadata.csv")
+
+# ### metadata OCF only
+
+filename = f"{dir}/system_metadata_OCF_ONLY.csv"
+
+passive_metadata_ocf = pd.read_csv(filename)
+
+passive_metadata_ocf["system_id"] = passive_metadata_ocf["ss_id"]
```
JackKelly, line 39: Instead of duplicating the columns, I'd maybe recommend using something like:

```python
passive_metadata_ocf = passive_metadata_ocf.rename(
    columns={
        "ss_id": "system_id",
        ...etc...
    }
)
```

scripts/passiv/passive.py
```diff
@@ -0,0 +1,32 @@
+""" Test script, to se if passive data works"""
```
JackKelly, line 1: Tiny typo

```suggestion
""" Test script, to see if passive data works"""
```

JackKelly (approved):
This all looks good!  Thank you!

How much extra work would it be to use the 2-minutely data (after pivoting it and resampling to 5-minutely?) so that we can have data that goes further back in time?  My thinking is that it would be great to be able to extend our dataset back as far as possible, both so we can use this new PV data right now in `predict_pv_yield.py` and because it would be able to train on pre-2020 ASAP in January (and maybe even during a potential hyper-param sweep over Christmas?)

Two general request:  Please add type hints to all the function signatures in `passiv_data.py`

Also, I'd maybe suggest making the `passive.py` and `passiv_data.py` filenames a bit longer and more descriptive?  Maybe `passive.py` could be called something like `test_converted_passivsystems_data.py`; and `passiv_data.py` could be called something like `convert_passivsystems_pv_data.py`?

nowcasting_dataset/data_sources/pv/pv_data_source.py
peterdudfield: oh that is much better - good idea

scripts/passiv/passiv_data.py
peterdudfield: ah yea Ill add a comment. Its bascially becasue 2021-Nov and 2021-Dec isnt there, but I didnt want it to break becasue of that

scripts/passiv/passiv_metadata.py
peterdudfield: good idea

peterdudfield:
> This all looks good! Thank you!
> 
> How much extra work would it be to use the 2-minutely data (after pivoting it and resampling to 5-minutely?) so that we can have data that goes further back in time? My thinking is that it would be great to be able to extend our dataset back as far as possible, both so we can use this new PV data right now in `predict_pv_yield.py` and because it would be able to train on pre-2020 ASAP in January (and maybe even during a potential hyper-param sweep over Christmas?)
> 
> Two general request: Please add type hints to all the function signatures in `passiv_data.py`
> 
> Also, I'd maybe suggest making the `passive.py` and `passiv_data.py` filenames a bit longer and more descriptive? Maybe `passive.py` could be called something like `test_converted_passivsystems_data.py`; and `passiv_data.py` could be called something like `convert_passivsystems_pv_data.py`?

Thanks for PR review

I agree would be good to do pre-2020 data as soon as possible. I'm a little be wary of doing this resampling, so I might speak to Jamie first. He may be able to do it for us, or provide the code he did it with. Both methods perhaps reduce the change of errors

JackKelly:
> I agree would be good to do pre-2020 data as soon as possible. I'm a little be wary of doing this resampling, so I might speak to Jamie first.

Good idea!

PR #417, opened 2021-11-17
feature/373 sun 10

### LINKED ISSUE #373: Only use timesteps where Sun's elevation is > 10 degrees above horizon
Throw away timesteps where, for a given geo region of interest, any geo location in that ROI has a Sun elevation <= 10 degrees.

### LINKED ISSUE #416: strange Sun location
**Describe the bug**
One sun location systems is at 
lat lon: 30,102

This is not in the UK

**To Reproduce**
Run 
```
zarr_path='tests/data/sun/test.zarr'
    sun_data_source = SunDataSource(zarr_path=zarr_path, history_minutes=30, forecast_minutes=60)
print(sun_data_source.elevation.iloc[:,382:384]
```

**Expected behavior**
locations systems to be in the UK

peterdudfield:
# Pull Request

## Description

- filter sun elevation data if >10
- taken a simple approach where we just look at the elevation at London at that time

Fixes #416 

## How Has This Been Tested?

- usual unittests
- print statements show lat and lons are now in the UK

- [ ] No
- [x] Yes

## Checklist:


nowcasting_dataset/data_sources/sun/sun_data_source.py
```diff
@@ -98,10 +99,26 @@ def get_locations(self, t0_datetimes: pd.DatetimeIndex) -> Tuple[List[Number], L
...
     def datetime_index(self):
-        """The datetime index of this datasource"""
-        raise NotImplementedError(
-            "Sun data should not be used for datetime_index. "
-            "This is because normally the data is available all the time, "
-            "except for when using test data. "
-            "This is becasue data from 2019 is extrapolate on to other years. "
+        """Get datetimes where elevation > 10"""
+
+        # london
```
peterdudfield, line 104: expand on commnet

nowcasting_dataset/data_sources/sun/sun_data_source.py
```diff
@@ -98,10 +99,26 @@ def get_locations(self, t0_datetimes: pd.DatetimeIndex) -> Tuple[List[Number], L
         raise NotImplementedError("Sun data should not be used to get batch locations")
 
     def datetime_index(self):
-        """The datetime index of this datasource"""
-        raise NotImplementedError(
-            "Sun data should not be used for datetime_index. "
-            "This is because normally the data is available all the time, "
-            "except for when using test data. "
-            "This is becasue data from 2019 is extrapolate on to other years. "
+        """Get datetimes where elevation > 10"""
```
peterdudfield, line 102: >=

JackKelly (approved):
LGTM!  Thanks!

Just to check:  Does `self.elevation.index` span from 2020-01-01 to 2021-08-31?

peterdudfield:
sun data is calculate on 2019, but then 

> LGTM! Thanks!
> 
> Just to check: Does `self.elevation.index` span from 2020-01-01 to 2021-08-31?

Oh thats a very good point. Sun data is actually from 2019 but then assumed it correct for all years. So ill need to add something for that. 

[ref](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/data_sources/sun/sun_data_source.py#L52)

JackKelly:
>  Sun data is actually from 2019 but then assumed it correct for all years. So ill need to add something for that.

I've been wondering for a while if we should have a config option for "start_date" and "end_date" for the entire dataset.  e.g. we'd set start_date=2020-01-01 and end_date=2021-08-31.  That might speed up processing (e.g. tell PVDataSource to only load data from start_date to end_date).  And would enable SunDataSource to "copy" its one year of Sun data to fill start_date to end_date.  Does that sound like an OK solution?

peterdudfield:
> > Sun data is actually from 2019 but then assumed it correct for all years. So ill need to add something for that.
> 
> I've been wondering for a while if we should have a config option for "start_date" and "end_date" for the entire dataset. e.g. we'd set start_date=2020-01-01 and end_date=2021-08-31. That might speed up processing (e.g. tell PVDataSource to only load data from start_date to end_date). And would enable SunDataSource to "copy" its one year of Sun data to fill start_date to end_date. Does that sound like an OK solution?

I think that would be a good thing to do. I guess most of the time the zarr data will about match all the other data. But yea I think that isnt too much work and would be fairly easily to do?

JackKelly:
Sounds good.  I've started a new issue about implementing `start_date` and `end_date` in the config yaml: #425

PR #477, opened 2021-11-19
Issue/pv gsp normalisation

peterdudfield:
# Pull Request

## Description

- add pv and gsp normalization variable
- update capacity in fake data
- update tests

Fixes #

## How Has This Been Tested?

- normal unitests
- added unittest for normalization

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/fake.py
+++ b/nowcasting_dataset/data_sources/fake.py
@@ -260,18 +260,7 @@ def create_gsp_pv_dataset(
         coords=coords,
     )  # Fake data for testing!
 
-    if time_dependent_capacity:
-        capacity = xr.DataArray(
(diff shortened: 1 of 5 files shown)


jacobbieker (approved):
Looks good! But aren't we trying to do all the normalization in dataloader? so shouldn't this be moved from here to there?

peterdudfield:
> Looks good! But aren't we trying to do all the normalization in dataloader? so shouldn't this be moved from here to there?

Good point 

Although - basically this is a useful function for plotting. Perhaps the difference here is we are not overwriting anything by normalization, just a function where we can get the normalized power

jacobbieker:
Ah okay, sounds good!

PR #480, opened 2021-11-19
remove metadata load from batch

### LINKED ISSUE #368: Remove `MetadataDataSource` as all data  is contained in csv
## Detailed Description
Apparently MetadataDAtaSource isn't being written by the new Manager, and instead metadata is only used in the fake Batch class. So removing it should not have any bad effects. 

## Context
This came up in #314 as I assumed that the metadata files were written too, but it turns out they are not anymore, so instead the t0 times should be loaded from the CSV file.

peterdudfield:
# Pull Request

## Description

Remove metadata objects from batch
make sure objects are made in batch when loading

Fixes #368 - half of

## How Has This Been Tested?

unitests

- [ ] No
- [x] Yes

## Checklist:


jacobbieker (approved):
LGTM!

PR #506, opened 2021-11-24
compute normalisation stats from training set

### LINKED ISSUE #499: Compute means and standard deviations using the `train` set
## Detailed Description
At the moment, we're computing the means and standard deviations across each entire source dataset.

Instead, I think we probably want to be computing the means and standard deviations across _just_ the pre-prepared training set.  This has several advantages:

- It's more representative of the data the ML model will actually see during training.  For example, we're excluding nighttime hours in our training set but, at the moment, our means and standard deviations are computed using nighttime hours.  In the training dataset, we're only using data over the landmass of the UK.  But our means and stds are computed over the land and ocean and France.
- It's less data :slightly_smiling_face: 
- It's a good opportunity to check (again!) the validity of the on-disk batches
- It prevents "leaking" information for the test set into the training procedure

What do you guys think?

I'm happy to take a shot at this later today...

JackKelly:
# Pull Request

## Description

Fixes #499

## How Has This Been Tested?

- [ ] No
- [x] Yes

## Checklist:


JackKelly:
this is just a very simple (and pretty scruffy notebook)... I'll make an issue to turn this notebook into a "proper" script another time... but that's not a priority now; this notebook has done what it needs to do...

PR #523, opened 2021-12-03
add start and end date

### LINKED ISSUE #425: Implement a `start_date` and `end_date` setting in the config YAML
have a config option for "start_date" and "end_date" for the entire dataset. e.g. we'd set start_date=2020-01-01 and end_date=2021-08-31. That might speed up processing (e.g. tell PVDataSource to only load data from start_date to end_date). And would enable SunDataSource to "copy" its one year of Sun data to fill start_date to end_date. Does that sound like an OK solution?

The `start_date` and `end_date` would be passed to every `DataSource` before the `DataSource` calls `DataSource.open()`

Related to: https://github.com/openclimatefix/nowcasting_dataset/pull/417#issuecomment-971522474

peterdudfield:
# Pull Request

## Description

add configuration for start and end datetimes for PV and GSP models

Fixes #425 

## How Has This Been Tested?

normal unittests

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/config/model.py
+++ b/nowcasting_dataset/config/model.py
@@ -11,6 +11,7 @@ Separate Pydantic models in
 are used to validate the values of the data itself.
 
 """
+import logging
 from datetime import datetime
 from pathlib import Path
 from typing import Optional, Union
@@ -31,6 +32,8 @@ from nowcasting_dataset.dataset.split import split
 
 IMAGE_SIZE_PIXELS_FIELD = Field(64, description="The number of pixels of the region of interest.")
 METERS_PER_PIXEL_FIELD = Field(2000, description="The number of meters per pixel.")
+
+logger = logging.getLogger(__name__)
 
 
 class General(BaseModel):
@@ -309,6 +312,19 @@ class InputData(BaseModel):
         Run through the PV and GSP data sources and  if the start or end date are not set,
         then set them to this values
         """
+
+        start_datetime = values["start_datetime"]
+        end_datetime = values["end_datetime"]
+
+        # check start datetime is less than end datetime
+        if start_datetime >= end_datetime:
+            message = (
+                f"Start datetime ({start_datetime}) "
+                f"should be less than end datetime ({end_datetime})"
+            )
+            logger.error(message)
+            assert Exception(message)
+
         # It would be much better to use nowcasting_dataset.data_sources.ALL_DATA_SOURCE_NAMES,
         # but that causes a circular import.
         ALL_DATA_SOURCE_NAMES = (
@@ -323,10 +339,10 @@ class InputData(BaseModel):
 
         for data_source_name in enabled_data_sources:
             if values[data_source_name].start_datetime is None:
-                values[data_source_name].start_datetime = values["start_datetime"]
+                values[data_source_name].start_datetime = start_datetime
 
             if values[data_source_name].end_datetime is None:
-                values[data_source_name].end_datetime = values["end_datetime"]
+                values[data_source_name].end_datetime = end_datetime
 
         return values
 
--- a/nowcasting_dataset/config/model.py
+++ b/nowcasting_dataset/config/model.py
@@ -175,6 +175,17 @@ class GSP(DataSourceMixin):
     gsp_image_size_pixels: int = IMAGE_SIZE_PIXELS_FIELD
     gsp_meters_per_pixel: int = METERS_PER_PIXEL_FIELD
 
+    start_date: datetime = Field(
+        None,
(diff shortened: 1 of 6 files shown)


nowcasting_dataset/config/model.py
```diff
@@ -86,7 +86,22 @@ def seq_length_60_minutes(self):
...
+
+    start_datetime: datetime = Field(
+        None,
+        description="Load date from data sources from this date. "
+        "If None, this will get overwritten by InputData.start_date. ",
+    )
+    end_datetime: datetime = Field(
+        None,
+        description="Load date from data sources up to this date. "
+        "If None, this will get overwritten by InputData.start_date. ",
```
peterdudfield, line 100: end date

nowcasting_dataset/config/model.py
```diff
@@ -240,6 +255,13 @@ class InputData(BaseModel):
         ),
     )
 
+    start_datetime: datetime = Field(
```
JackKelly, line 258: Is this needed, given that the `StartEndDatetimeMixin` will inject a `start_datetime` and `end_datetime` fields?

nowcasting_dataset/config/model.py
```diff
@@ -279,6 +301,35 @@ def set_forecast_and_history_minutes(cls, values):
...
 
+    @root_validator
+    def set_start_and_end_datetime(cls, values):
+        """
+        Set start and end date, if needed.
+
+        Run through the PV and GSP data sources and  if the start or end date are not set,
+        then set them to this values
+        """
+        # It would be much better to use nowcasting_dataset.data_sources.ALL_DATA_SOURCE_NAMES,
```
JackKelly, line 312: might this function also be a good place to check that `end_datetime > start_datetime`?

nowcasting_dataset/data_sources/gsp/gsp_data_source.py
```diff
@@ -41,11 +41,9 @@ class GSPDataSource(ImageDataSource):
     # zarr_path of where the gsp data is stored
     zarr_path: Union[str, Path]
     # start datetime, this can be None
-    # TODO: Issue #425: Use config to set start_dt and end_dt.
-    start_dt: Optional[datetime] = pd.Timestamp("2020-01-01")
+    start_datetime: Optional[datetime] = pd.Timestamp("2020-01-01")
```
JackKelly, line 44: I'd be very tempted to remove these defaults.  Defaults like this can, all too easily, mean that the code silently does the wrong thing!  Or to set the default to `None` (if `None` means "use the extent of the dataset")

JackKelly (approved):
looks good! thanks!

please also remove the `logger.warning("{GSP,PV}DataSource is using hard-coding start_dt and end_dt")` on:
* line 70 of gsp_data_source.py
* line 52 of pv_data_source.py

nowcasting_dataset/data_sources/gsp/gsp_data_source.py
peterdudfield: Yea I can put them as None

nowcasting_dataset/config/model.py
peterdudfield: I was thinking that these are the general or default ones, and then in PV and GSP, separate ones can be set

JackKelly: Actually, that raises an interesting question:

What's the use-case where we'd want differente start_datetimes and end_datetimes for each datasource?  Or should start_datetime and end_datetime always be "global"

peterdudfield: Perhaps if one data source is dodge before a certain date, but we still want data from before this date for the batches, but with this data source missing. Perhaps this is a bit far fetched ...

JackKelly: That's a good point about avoiding dodgy data...

But `Manager` starts by computing the intersection of all the t0_datetimes available across all the DataSources.  So if one data source starts really late, then all the data sources will start really late.

so, yeah, I'm starting to wonder if maybe we should keep things simple and just have a single, global `start_datetime` and `end_datetime`?  What do you think?

peterdudfield: Maybe its simplier to remove this bit - simplier is often better

peterdudfield: Yea two different ways to do it
1. Currently its just a mxinin in PV and GSP
2. other option is to have a global one which is move into PV GSP through a validation method. 

I think 1 is simple pydantic model. Can always change this in the future if we find we are changing things.

PR #536, opened 2021-12-10
extend NWP forecasts by 1 hour

JackKelly:
# Pull Request

## Description

Very simple PR!  Just adding 1 hour to the NWP forecasts.

Otherwise, at the moment, if the GSP PV data ends on a half-hour, then the NWP forecast ends 30 minutes before the GSP "forecast":

Here's some example data plotted.  The green line is NWP irradiance.  The orange line is GSP PV (just for the forecast horizon) 
 If you look at the top-left subplot, you can see that the NWP data (green line) ends half an hour early..

image


peterdudfield (approved):
do you want to add GSP,PV history too?

JackKelly:
good idea!  Done!

PR #540, opened 2021-12-10
prepare larger satellite and pv batches to slot into v15

### LINKED ISSUE #529: Create new pre-prepared batches with larger regions of interest
## Detailed Description
Set the `image_size_pixels = 102` and `meters_per_pixel = 4000` for `satellite` and `pv`.

## Context
After working with optical flow for a bit, it's really clear that, in order to predict 2 hours into the future, we need to use much larger satellite images (so that we can see clouds 2 hours ahead of time).

We probably don't need such large regions of interest for NWPs.

## Possible Implementation
Instead of creating a `v17` so soon after creating `v16`, maybe we could just change `on_premises.yaml` and run `prepare_ml_data.py --data_source pv --data_source satellite --overwrite_batches`?

JackKelly:
# Pull Request

## Description

Slightly hacky for now... creating new PV and satellite batches with larger regions of interest so that we can see if we can get our ML models to associate clouds with PV power drops in the recent history.

The output_data.filepath is set to v17.  But I've manually copied the `spatial_and_temporal_locations_of_each_example.csv` files from `v15/train/` and `v15/test/` into a new `v17/` path, and I'm just running `prepare_ml_data.py --data_source pv --data_source satellite` for now (to avoid having to re-create everything).

Fixes #529

## How Has This Been Tested?

- [ ] No
- [x] Yes, `prepare_ml_data.py` runs

## Checklist:

### DIFF
--- a/nowcasting_dataset/config/on_premises.yaml
+++ b/nowcasting_dataset/config/on_premises.yaml
@@ -24,15 +24,15 @@ input_data:
       - hcc
     nwp_image_size_pixels: 64
     nwp_zarr_path: /mnt/storage_ssd_8tb/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/NWP/UK_Met_Office/UKV/zarr/UKV_intermediate_version_3.zarr
-    # nwp_forecast_minutes: 180
-    nwp_history_minutes: 60
+    # forecast_minutes: 180
+    history_minutes: 60
 
   #---------------------- PV -------------------
   pv:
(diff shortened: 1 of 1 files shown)


peterdudfield:
This is probably not needed now as we have https://github.com/openclimatefix/nowcasting_dataset/pull/647. What do you think @JackKelly

JackKelly:
Completely agree! Let's close this!

JackKelly:
Thanks for helping to tidy up, BTW!

PR #555, opened 2021-12-20
Issue/dl 109 load 2 batches

peterdudfield:
# Pull Request

## Description

two functions taken from dataloader but only relate to Batch
- download and load batch
- join two batches together

Fixes https://github.com/openclimatefix/nowcasting_dataloader/issues/109

## How Has This Been Tested?

normal unittests
- added extra unittests
- 
- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/dataset/batch.py
+++ b/nowcasting_dataset/dataset/batch.py
@@ -259,9 +259,9 @@ def join_two_batches(
     batch = batches[0]
     batch_size = batch.metadata.batch_size
 
+    # create random indexes, if needed
     if first_batch_examples is None:
         first_batch_examples = np.random.randint(low=0, high=batch_size, size=int(batch_size / 2))
-
     second_batch_examples = [i for i in range(0, batch_size) if i not in first_batch_examples]
 
     for data_source in data_sources_names:
@@ -279,6 +279,7 @@ def join_two_batches(
 
     # merge metadata
     metadata = batch.metadata
+    # loop over metadata keys, but no 'batch_size'
     for metadata_key in Metadata.__fields__.keys():
         if metadata_key != "batch_size":
             data = np.array(getattr(batch.metadata, metadata_key))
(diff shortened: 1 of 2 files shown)


nowcasting_dataset/dataset/batch.py
```diff
@@ -183,3 +229,62 @@ def data_sources(self):
...
+
+    assert len(batches) == 2, f"Can only join list of two batches, was given {len(batches)} batches"
+
+    if data_sources_names is None:
+        data_sources_names = list(Example.__fields__.keys())
+
+    batch = batches[0]
+    batch_size = batch.metadata.batch_size
+
+    if first_batch_examples is None:
```
peterdudfield, line 262: add comment

nowcasting_dataset/dataset/batch.py
```diff
@@ -183,3 +229,63 @@ def data_sources(self):
...
+    if data_sources_names is None:
+        data_sources_names = list(Example.__fields__.keys())
+
+    batch = batches[0]
+    batch_size = batch.metadata.batch_size
+
+    # create random indexes, if needed
+    if first_batch_examples is None:
+        first_batch_examples = np.random.randint(low=0, high=batch_size, size=int(batch_size / 2))
+    second_batch_examples = [i for i in range(0, batch_size) if i not in first_batch_examples]
```
jacobbieker, line 265: If we want them to be random, why not have random examples from both? As in, as it is here, if the 0th and 2nd examples are chosen from the first batch, they can't be chosen from the second. For it to be more random, it could allow for having the 0th and 2nd examples of the first and second batch in one batch together, if that makes sense! I get it might make it a bit more complicated as we then might not be able to concat along the example?

jacobbieker (approved):
Looks good to me! Thanks! Just one smallish comment on how random we want to be with the batches

nowcasting_dataset/dataset/batch.py
peterdudfield: That is a very good point. That would make it more random, but the code would be a little be more complicated as you said. 

ill have a think whats the best thing to do for the long-term

peterdudfield: I went with your idea - thanks @jacobbieker - makes it all a bit more random

PR #123, opened 2021-09-16
add pre commit hook

peterdudfield:
# Pull Request

## Description

Add pre commit hooks  - https://pre-commit.com/
This means 'blacks' will run just before everytime someone makes a commit

Fixes issue #

## How Has This Been Tested?

Similar usage on nowcasting_utils

- [ ] No
- [x] Yes

## Checklist:

- [x] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/nowcasting/blob/main/coding_style.md)
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked my code and corrected any misspellings

PR #562, opened 2021-12-21
first try and making multiple input data files for pv

peterdudfield:
# Pull Request

## Description

PV data can have two datasources

Fixes #

## How Has This Been Tested?

- [ ] No
- [ ] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/config/model.py
+++ b/nowcasting_dataset/config/model.py
@@ -155,13 +155,13 @@ class StartEndDatetimeMixin(Base):
 class PV(DataSourceMixin, StartEndDatetimeMixin):
     """PV configuration model"""
 
-    pv_filename: str = Field(
-        "gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_timeseries_batch.nc",
-        description=("The NetCDF file holding the solar PV power timeseries."),
+    pv_filenames: List = Field(
+        ["gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_timeseries_batch.nc"],
+        description=("List of the NetCDF file holding the solar PV power timeseries."),
     )
-    pv_metadata_filename: str = Field(
+    pv_metadata_filenames: List = Field(
         "gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_metadata.csv",
-        description="The CSV file describing each PV system.",
+        description="List of the CSV file describing each PV system.",
     )
     n_pv_systems_per_example: int = Field(
         DEFAULT_N_PV_SYSTEMS_PER_EXAMPLE,
--- a/nowcasting_dataset/config/on_premises.yaml
+++ b/nowcasting_dataset/config/on_premises.yaml
@@ -29,8 +29,10 @@ input_data:
 
   #---------------------- PV -------------------
   pv:
-    pv_filename: /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/passiv.netcdf
-    pv_metadata_filename: /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/system_metadata_OCF_ONLY.csv
+    pv_filenames:
+      - /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/passiv.netcdf
+    pv_metadata_filenames:
+      - /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/system_metadata_OCF_ONLY.csv
     get_center: false
     history_minutes: 90
 
--- a/nowcasting_dataset/data_sources/pv/pv_data_source.py
+++ b/nowcasting_dataset/data_sources/pv/pv_data_source.py
@@ -32,8 +32,8 @@ class PVDataSource(ImageDataSource):
     defined by image_size_pixels and meters_per_pixel.
     """
 
-    filename: Union[str, Path]
-    metadata_filename: Union[str, Path]
+    filenames: Union[str, Path, List]
+    metadata_filenames: Union[str, Path, List]
     # TODO: Issue #425: Use config to set start_dt and end_dt.
     start_datetime: Optional[datetime.datetime] = None
     end_datetime: Optional[datetime.datetime] = None
@@ -49,12 +49,18 @@ class PVDataSource(ImageDataSource):
         """Post Init"""
         super().__post_init__(image_size_pixels, meters_per_pixel)
 
+        if not isinstance(self.filenames, list):
+            self.filenames = [self.filenames]
+
+        if not isinstance(self.metadata_filenames, list):
+            self.metadata_filenames = [self.metadata_filenames]
+
         self.rng = np.random.default_rng()
         self.load()
 
     def check_input_paths_exist(self) -> None:
         """Check input paths exist.  If not, raise a FileNotFoundError."""
-        for filename in [self.filename, self.metadata_filename]:
+        for filename in self.filenames + self.metadata_filenames:
             nd_fs_utils.check_path_exists(filename)
 
     def load(self):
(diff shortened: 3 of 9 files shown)


nowcasting_dataset/config/model.py
```diff
@@ -161,13 +161,13 @@ def check_start_and_end_datetime(cls, values):
 class PV(DataSourceMixin, StartEndDatetimeMixin):
     """PV configuration model"""
 
-    pv_filename: str = Field(
-        "gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_timeseries_batch.nc",
-        description=("The NetCDF file holding the solar PV power timeseries."),
+    pv_filenames: List = Field(
```
JackKelly, line 164: tiny tweak: [As of Python 3.9, we can use `list` and `tuple` instead of `typing.List` and `typing.Tuple`](https://stackoverflow.com/a/39458225/732596).

```suggestion
    pv_filenames: list = Field(
```

nowcasting_dataset/config/model.py
```diff
@@ -161,13 +161,13 @@ def check_start_and_end_datetime(cls, values):
...
+        ["gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_timeseries_batch.nc"],
+        description=("List of the NetCDF file holding the solar PV power timeseries."),
```
JackKelly, line 166: Tiny typo :slightly_smiling_face: 

```suggestion
        description=("List of the NetCDF files holding the solar PV power timeseries."),
```

nowcasting_dataset/config/model.py
```diff
@@ -161,13 +161,13 @@ def check_start_and_end_datetime(cls, values):
...
     )
-    pv_metadata_filename: str = Field(
+    pv_metadata_filenames: List = Field(
         "gs://solar-pv-nowcasting-data/PV/PVOutput.org/UK_PV_metadata.csv",
-        description="The CSV file describing each PV system.",
+        description="List of the CSV file describing each PV system.",
```
JackKelly, line 170:
```suggestion
        description="List of the CSV files describing each PV system.",
```

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -47,14 +47,20 @@ class PVDataSource(ImageDataSource):
 
     def __post_init__(self, image_size_pixels: int, meters_per_pixel: int):
         """Post Init"""
+        if not isinstance(self.filenames, list):
```
JackKelly, line 50: In general, I'd recommend that we allow `filenames` to be _either_ a `list` or a `tuple`.

```suggestion
        if not isinstance(self.filenames, (tuple, list)):
```

Also, to keep the code as simple as possible, I'd maybe recommend that `PVDataSource` simply blows up if `self.filenames` isn't a `tuple` or a `list`!  (Are there any use-cases where `self.filenames` might still be a `str`?)

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -98,11 +119,34 @@ def _load_metadata(self):
 
     def _load_pv_power(self):
 
-        logger.debug(f"Loading PV Power data from {self.filename}")
+        logger.debug(f"Loading PV Power data from {self.filenames}")
 
-        pv_power = load_solar_pv_data(
-            self.filename, start_dt=self.start_datetime, end_dt=self.end_datetime
-        )
+        # collect all metadata together
```
JackKelly, line 124:
```suggestion
        # collect all PV power timeseries together
```

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -72,9 +78,24 @@ def get_data_model_for_batch():
...
+        pv_metadata = []
+        for i, metadata_filename in enumerate(self.metadata_filenames):
+
+            # read metadata file
+            metadata = pd.read_csv(metadata_filename, index_col="system_id")
 
-        pv_metadata = pd.read_csv(self.metadata_filename, index_col="system_id")
+            # add i to index, to make sure the indexes are unique
+            new_index = metadata.index.astype(str)
+            new_index = new_index + f"_{i}"
```
JackKelly, line 92: Good thinking about ensuring the PV system IDs remain unique!

However, a shortcoming of using `<pv id>_<i>`  is that `i` depends on the (arbitrary) order of the filenames in the config YAML. So, when the code manually cleans some specific PV systems (line 152), the code can't unambiguously identify, for example, "PVOutput.org system #123".

Is there a way to guarantee that, for example, PVOutput.org systems will always have the suffix `0` and PassivSystem PV systems will always have the suffix `1`?

Or maybe use a [Pandas MultiIndex](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html), where one level of the multi-index is `"PVOutput.org"` or `"PassivSystems"`, and the other level is the integer PV system ID?  (But I don't know if the downstream code will like using MultiIndexes!)

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -98,11 +119,34 @@ def _load_metadata(self):
...
+            pv_power.columns = new_columns
+
+            pv_power_all.append(pv_power)
+
+        # join together pv power. Index is time, columns are pv ids. Make sure the ids are unique
+        pv_power = pv_power_all.pop(0)
+        while len(pv_power_all) > 0:
+
+            # Make sure the ids are unique
+            indexes = list(pv_power.columns)
```
JackKelly, line 145: Maybe rename `indexes` and `new_indexes` to `pv_system_ids` and `new_pv_system_ids`?

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -98,11 +119,34 @@ def _load_metadata(self):
...
+
+            # add i to columns, to make sure the columns are unique
+            new_columns = pv_power.columns.astype(str)
+            new_columns = new_columns + f"_{i}"
+            pv_power.columns = new_columns
+
+            pv_power_all.append(pv_power)
+
+        # join together pv power. Index is time, columns are pv ids. Make sure the ids are unique
+        pv_power = pv_power_all.pop(0)
```
JackKelly, line 141: Sorry if I've misunderstood :slightly_smiling_face: but could lines 141 to 149 be replaced with something like:

```python
pv_power = pd.concat(pv_power_all, axis="columns")
assert not pv_power.columns.duplicated().any()
```

nowcasting_dataset/data_sources/pv/pv_data_source.py
```diff
@@ -118,6 +162,7 @@ def _load_pv_power(self):
         # Resample to 5-minutely and interpolate up to 15 minutes ahead.
         # TODO: Issue #301: Give users the option to NOT resample (because Perceiver IO
         # doesn't need all the data to be perfectly aligned).
+        print(pv_power)
```
JackKelly, line 165: Remove this `print`?

nowcasting_dataset/config/on_premises.yaml
```diff
@@ -30,8 +30,10 @@ input_data:
 
   #---------------------- PV -------------------
   pv:
-    pv_filename: /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/passiv.netcdf
-    pv_metadata_filename: /mnt/storage_b/data/ocf/solar_pv_nowcasting/nowcasting_dataset_pipeline/PV/Passiv/ocf_formatted/v0/system_metadata_OCF_ONLY.csv
+    pv_filenames:
```
JackKelly, line 33: Could we add the `PVOutput.org` data to the `on_premises.yml` config file?

JackKelly (approved):
Looks good! A few comments. Nothing major :slightly_smiling_face: 

Thanks loads for tackling this - it'll be great to have all this lovely PV data for training!

nowcasting_dataset/data_sources/pv/pv_data_source.py
peterdudfield: we could order the PV filenames list by alphabetic order, then it should always be one way round. We could do this in the pydantic model. Have to make sure the metdata is ordered the same way

nowcasting_dataset/data_sources/pv/pv_data_source.py
peterdudfield: yes, i think that is a better way, let me try it

nowcasting_dataset/data_sources/pv/pv_data_source.py
JackKelly: Hmm, I gotta admit that feels fragile. What happens if we only use PVOutput.org data (which, when using both PVOutput.org and PassivSystems data would be the second entry. But would be the first (and only) entry if we only use PVOutput.org data)

How about something like detecting if the string "pvoutput" or "passiv" is in the `filename.lower()` string, and then appending either "pvoutput" or "passiv" to every PV system ID?

peterdudfield: Yea perhaps the model should be something like this
pv_files: [
- pv_filename:
- pv_metedata
- label
]

Then we can use the label to do the new ids. We could even add a validator to make sure all the lables are different

JackKelly: SGTM!

peterdudfield: It certainly can!

peterdudfield: I've tidied up the config model, as described above. 

I then used an encode (full code [here](https://github.com/openclimatefix/nowcasting_dataset/blob/issue/554-both-pv/nowcasting_dataset/data_sources/pv/pv_data_source.py#L471)):
```
label_index = PV_PROVIDERS.index(label)
new_index = [str(int(col) * 10 + label_index) for col in indexes]
```

I've added an assert if the list of pv provers ever is more than 10 (i don't think this will happen)

PR #565, opened 2021-12-22
Issue/539 more pv data

### LINKED ISSUE #539: Include summary stats for PV power (to capture info when there's > 128 PV systems in ROI)
## Detailed Description
Our batches only store up to 128 PV systems per example.  But sometimes there are more PV systems available.  At the moment, those systems get lost.  But maybe we should always compute some summary stats (mean, std, min, max?) across all PV systems, and store those stats, so our models can always benefit at least a bit from all the PV systems in the region of interest.

peterdudfield:
# Pull Request

## Description

- Increase number of pv systems in example to 2048
- order pv systems by how close they are to the center of the examples
- NOT: provide summary statistics. This is perhaps something we could do in the dataloader

Fixes #539 

## How Has This Been Tested?

- normal unittests
- added unittest

- [ ] No
- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/consts.py
+++ b/nowcasting_dataset/consts.py
@@ -24,7 +24,7 @@ SUN_ELEVATION_ANGLE = "sun_elevation_angle"
 
 PV_YIELD = "pv_yield"
 PV_DATETIME_INDEX = "pv_datetime_index"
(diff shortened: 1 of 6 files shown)


jacobbieker (approved):
Looks good!

PR #587, opened 2022-01-27
Rename time_resolution_minutes to sample_period_minutes to make the code consistent 

### LINKED ISSUE #584: Rename `time_resolution_minutes` to `sample_period_minutes` to make the code consistent
## Detailed Description
`model.py` uses `time_resolution_minutes`.  Elsewhere in the code this same concept is called `sample_period_minutes`.

## Context
Good to be consistent throughout the code.

RishiKumarRay:
# Pull Request

## Description
Rename time_resolution_minutes to sample_period_minutes to make the code consistent

Fixes #584

## Checklist:

- [ x] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/nowcasting/blob/main/coding_style.md)
- [ x] I have performed a self-review of my own code
- [ x] I have made corresponding changes to the documentation
- [ ]x I have added tests that prove my fix is effective or that my feature works
- [ x] I have checked my code and corrected any misspellings

### DIFF
--- a/nowcasting_dataset/data_sources/optical_flow/optical_flow_data_source.py
+++ b/nowcasting_dataset/data_sources/optical_flow/optical_flow_data_source.py
@@ -55,7 +55,7 @@ class OpticalFlowDataSource(DataSource):
         the input image after it has been "flowed".
     source_data_source_class_name: Either HRVSatelliteDataSource or SatelliteDataSource.
     channels: The satellite channels to compute optical flow for.
-    time_resolution_minutes: time resolution of optical flow images.
+    sample_period_minutes: time resolution of optical flow images.
         Needs to be the same as the satellite images used to make the flow fields.
     """
 
@@ -65,7 +65,7 @@ class OpticalFlowDataSource(DataSource):
     meters_per_pixel: int = 2000
     output_image_size_pixels: int = 32
     source_data_source_class_name: str = "SatelliteDataSource"
-    time_resolution_minutes: int = 15
+    sample_period_minutes: int = 15
 
     def __post_init__(self):  # noqa
         assert self.output_image_size_pixels <= self.input_image_size_pixels, (
@@ -98,7 +98,7 @@ class OpticalFlowDataSource(DataSource):
     @property
     def sample_period_minutes(self) -> int:
         """Override the default sample minutes"""
-        return self.time_resolution_minutes
+        return self.sample_period_minutes
 
     def open(self):
         """Open the underlying self.source_data_source."""
--- a/nowcasting_dataset/config/model.py
+++ b/nowcasting_dataset/config/model.py
@@ -126,13 +126,13 @@ class DataSourceMixin(Base):
 class TimeResolutionMixin(Base):
     """Time resolution mix in"""
 
-    time_resolution_minutes: int = Field(
+    sample_period_minutes: int = Field(
         15,
         description="The temporal resolution (in minutes) of the satellite images."
         "Note that this needs to be divisible by 5.",
     )
 
-    @validator("time_resolution_minutes")
+    @validator("sample_period_minutes")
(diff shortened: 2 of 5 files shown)


nowcasting_dataset/data_sources/optical_flow/optical_flow_data_source.py
```diff
@@ -98,7 +98,7 @@ def __post_init__(self):  # noqa
     @property
     def sample_period_minutes(self) -> int:
         """Override the default sample minutes"""
-        return self.time_resolution_minutes
+        return self.sample_period_minutes
```
JackKelly, line 101: Ah, sorry, I hadn't thought about this issue when I wrote issue #587!  I don't think Python allows classes where a `@property` method has the same name as an attribute of the class.  (Because, in Python, methods are attributes, too :slightly_smiling_face:).  Sorry, I should've thought about this!

Perhaps the cleanest solution is to completely get rid of the `sample_period_minutes` `@property` method from `DataSource` and all subclasses of `DataSource`. But then I think we bump into the issue that dataclass subclasses can't set new default values. (Although maybe that's changed in more recent versions of Python?)

If you're up for it, it'd be amazing if you could have a tinker and see if it's possible to remove the `sample_period_minutes @property` method, and then find a way to allow subclasses of `DataSource` to set their own default values for `sample_period_minutes`?

JackKelly (commented):
Thank you so much for jumping in and writing a PR for issue #587! (And welcome to the project!)

Just one comment...

JackKelly:
@all-contributors please add @RishiKumarRay for code

nowcasting_dataset/data_sources/optical_flow/optical_flow_data_source.py
RishiKumarRay: @JackKelly I have to learn python for this, I don't know much about python

JackKelly: OK, no rush!  It's totally up to you:  If you fancy doing this then please go for it, there's no rush for this issue!  But, if you'd prefer not to, then please just shout!

The relevant part of the Python docs might be [this page on inheritance in `dataclasses`](https://docs.python.org/3/library/dataclasses.html#inheritance).  On skim-reading the docs, it actually looks like it might be fine to just delete the `sample_period_minutes @property` method from all our code, and just use a `sample_period_minutes` attribute, which can be set to different values for each subclass.

RishiKumarRay: @JackKelly Sorry for this , I thought it was all about renaming thats why I raised the PR , maybe I am not suitable for this now to do the changes , If you want i will close this pull request and maybe you can assign someone else, i looked at the problem , removing sample_period_minutes @property method  is easy as you know but using just as atribute , I don't know much of it since i am totally unaware of python.

JackKelly: Cool, absolutely no worries!  Don't worry about deleting this PR, as your changes are definitely useful! Maybe someone else can pick up where this left off...  Thanks again for your help, and please do keep an eye out for other "renaming" issues!

RishiKumarRay: @JackKelly If it is useful , can you merge it , maybe some other issue should be opened for adding as attribute , Just a suggestion , Thanks :) also I will start learning python in my free time , thanks for suggestions

JackKelly: I'm not certain but I _think_ the code might break if we merge as is (even though the unittests pass... but I think the passing tests might be a "false pass" because maybe we don't have any unittests which call the `sample_period_minutes @property`?)

PR #592, opened 2022-01-28
Align fake batches in time.

peterdudfield:
# Pull Request

## Description

add an option to align fake batches in time. This is useful for the forecast service that tests on fake batches

## How Has This Been Tested?

add a unittest

- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/fake/batch.py
+++ b/nowcasting_dataset/data_sources/fake/batch.py
@@ -135,7 +135,7 @@ def gsp_fake(
     return GSP(xr_dataset)
 
 
-def metadata_fake(batch_size, temporally_align_examples: int = False) -> Metadata:
+def metadata_fake(batch_size, temporally_align_examples: bool = False) -> Metadata:
     """
     Make fake metadata object
 
--- a/nowcasting_dataset/data_sources/fake/batch.py
+++ b/nowcasting_dataset/data_sources/fake/batch.py
@@ -36,11 +36,21 @@ from nowcasting_dataset.dataset.xr_utils import (
 from nowcasting_dataset.geospatial import lat_lon_to_osgb
 
 
-def make_fake_batch(configuration: Configuration) -> dict:
-    """Make fake batch"""
+def make_fake_batch(configuration: Configuration, temporally_align_batches: bool = False) -> dict:
+    """
+    Make fake batch object
+
+    Args:
+        configuration: configuration of dataset
+        temporally_align_batches: option to align batches in time
+
+    Returns: dictionary of batch data
+    """
     batch_size = configuration.process.batch_size
 
-    metadata = metadata_fake(batch_size=batch_size)
+    metadata = metadata_fake(
+        batch_size=batch_size, temporally_align_batches=temporally_align_batches
+    )
 
     return dict(
         metadata=metadata,
@@ -125,8 +135,16 @@ def gsp_fake(
     return GSP(xr_dataset)
 
 
-def metadata_fake(batch_size):
-    """Make a xr dataset"""
+def metadata_fake(batch_size, temporally_align_batches: int = False) -> Metadata:
+    """
+    Make fake metadata object
+
+    Args:
+        batch_size: The size of the batch
+        temporally_align_batches: option to align batches in time
+
+    Returns: fake metadata
+    """
 
     # get random OSGB center in the UK
     lat = np.random.uniform(51, 55, batch_size)
@@ -134,7 +152,9 @@ def metadata_fake(batch_size):
     x_centers_osgb, y_centers_osgb = lat_lon_to_osgb(lat=lat, lon=lon)
 
     # get random times
-    t0_datetimes_utc = make_t0_datetimes_utc(batch_size)
+    t0_datetimes_utc = make_t0_datetimes_utc(
+        batch_size=batch_size, temporally_align_batches=temporally_align_batches
+    )
 
     metadata_dict = {}
     metadata_dict["batch_size"] = batch_size
--- a/nowcasting_dataset/data_sources/fake/batch.py
+++ b/nowcasting_dataset/data_sources/fake/batch.py
@@ -36,20 +36,20 @@ from nowcasting_dataset.dataset.xr_utils import (
 from nowcasting_dataset.geospatial import lat_lon_to_osgb
 
 
-def make_fake_batch(configuration: Configuration, temporally_align_batches: bool = False) -> dict:
+def make_fake_batch(configuration: Configuration, temporally_align_examples: bool = False) -> dict:
     """
     Make fake batch object
 
     Args:
         configuration: configuration of dataset
-        temporally_align_batches: option to align batches in time
+        temporally_align_examples: option to align examples (within the batch) in time
 
     Returns: dictionary of batch data
     """
     batch_size = configuration.process.batch_size
 
     metadata = metadata_fake(
-        batch_size=batch_size, temporally_align_batches=temporally_align_batches
+        batch_size=batch_size, temporally_align_examples=temporally_align_examples
     )
 
     return dict(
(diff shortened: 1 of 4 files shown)


jacobbieker (approved):

nowcasting_dataset/data_sources/fake/batch.py
```diff
@@ -36,11 +36,21 @@
...
 
-def make_fake_batch(configuration: Configuration) -> dict:
-    """Make fake batch"""
+def make_fake_batch(configuration: Configuration, temporally_align_batches: bool = False) -> dict:
+    """
+    Make fake batch object
+
+    Args:
+        configuration: configuration of dataset
+        temporally_align_batches: option to align batches in time
```
JackKelly, line 45: What's the use-case where we _don't_ want to align the batches in time?

nowcasting_dataset/data_sources/fake/batch.py
```diff
@@ -125,16 +135,26 @@ def gsp_fake(
     return GSP(xr_dataset)
 
 
-def metadata_fake(batch_size):
-    """Make a xr dataset"""
+def metadata_fake(batch_size, temporally_align_batches: int = False) -> Metadata:
```
JackKelly, line 138: tiny typo, I think?  :slightly_smiling_face: 

```suggestion
def metadata_fake(batch_size, temporally_align_batches: bool = False) -> Metadata:
```

nowcasting_dataset/data_sources/fake/utils.py
```diff
@@ -20,19 +20,25 @@ def join_list_data_array_to_batch_dataset(data_arrays: List[xr.DataArray]) -> xr
     return join_list_dataset_to_batch_dataset(datasets)
 
 
-def make_t0_datetimes_utc(batch_size):
+def make_t0_datetimes_utc(batch_size, temporally_align_batches: int = False):
```
JackKelly, line 23:
```suggestion
def make_t0_datetimes_utc(batch_size, temporally_align_batches: bool = False):
```

nowcasting_dataset/data_sources/fake/utils.py
```diff
@@ -20,19 +20,25 @@ def join_list_data_array_to_batch_dataset(data_arrays: List[xr.DataArray]) -> xr
...
     Returns: pandas index of t0 datetimes
     """
 
     all_datetimes = pd.date_range("2021-01-01", "2021-02-01", freq="5T")
-    t0_datetimes_utc = np.random.choice(all_datetimes, batch_size, replace=False)
+
+    if temporally_align_batches:
+        t0_datetimes_utc = list(np.random.choice(all_datetimes, 1)) * batch_size
+    else:
+        t0_datetimes_utc = np.random.choice(all_datetimes, batch_size, replace=False)
```
JackKelly, line 39: if we're sampling without replacement, then should we first check that `batch_size < len(all_datetimes)`?

nowcasting_dataset/data_sources/fake/utils.py
```diff
@@ -20,19 +20,25 @@ def join_list_data_array_to_batch_dataset(data_arrays: List[xr.DataArray]) -> xr
...
+        temporally_align_batches: option to align batches in time
 
     Returns: pandas index of t0 datetimes
     """
 
     all_datetimes = pd.date_range("2021-01-01", "2021-02-01", freq="5T")
-    t0_datetimes_utc = np.random.choice(all_datetimes, batch_size, replace=False)
+
+    if temporally_align_batches:
+        t0_datetimes_utc = list(np.random.choice(all_datetimes, 1)) * batch_size
```
JackKelly, line 37: Two quick thoughts:

I had to look up what the second positional argument to `choice` means! Please consider using something like `np.random.choice(all_datetimes, size=1)` to make it clear what the `1` means :slightly_smiling_face: 

Also, should we set the seed of the random number generator, so tests are repeatable? Using numpy's newish `Generator` class, so this function gets its own RNG which is guaranteed to create repeatable numbers?

JackKelly (approved):
Looks good! A few comments...

One thing that I didn't mention in the inline comments...

I wonder if we should _either_ change the name of the `temporally_align_batches` argument _or_ change the behaviour? :slightly_smiling_face: 

To me, "temporally aligning batches" suggests that we'd have a bunch of different `t0_datetimes`.  And that single list of `t0_datetimes` would be used across all modalities. So batch _i_ of satellite data would cover the same time period as batch _i_ of PV data, etc.  And batch _i_ would cover a different time period to batch _i+1_.

But, if I've understood correctly, if `temporally_align_batches` is set to `True` then the code repeatedly uses a single datetime. But the list of `t0_datetimes` could be different for each modality?  Is that right?

JackKelly (approved):
Looks good! A few comments...

One thing that I didn't mention in the inline comments...

I wonder if we should _either_ change the name of the `temporally_align_batches` argument _or_ change the behaviour? :slightly_smiling_face: 

To me, "temporally aligning batches" suggests that we'd have a bunch of different `t0_datetimes`.  And that single list of `t0_datetimes` would be used across all modalities. So batch _i_ of satellite data would cover the same time period as batch _i_ of PV data, etc.  And batch _i_ would cover a different time period to batch _i+1_.

But, if I've understood correctly, if `temporally_align_batches` is set to `True` then the code repeatedly uses a single datetime. But the list of `t0_datetimes` could be different for each modality?  Is that right?

peterdudfield:
> Looks good! A few comments...
> 
> One thing that I didn't mention in the inline comments...
> 
> I wonder if we should _either_ change the name of the `temporally_align_batches` argument _or_ change the behaviour? 🙂
> 
> To me, "temporally aligning batches" suggests that we'd have a bunch of different `t0_datetimes`. And that single list of `t0_datetimes` would be used across all modalities. So batch _i_ of satellite data would cover the same time period as batch _i_ of PV data, etc. And batch _i_ would cover a different time period to batch _i+1_.
> 
> But, if I've understood correctly, if `temporally_align_batches` is set to `True` then the code repeatedly uses a single datetime. But the list of `t0_datetimes` could be different for each modality? Is that right?



> Looks good! A few comments...
> 
> One thing that I didn't mention in the inline comments...
> 
> I wonder if we should _either_ change the name of the `temporally_align_batches` argument _or_ change the behaviour? 🙂
> 
> To me, "temporally aligning batches" suggests that we'd have a bunch of different `t0_datetimes`. And that single list of `t0_datetimes` would be used across all modalities. So batch _i_ of satellite data would cover the same time period as batch _i_ of PV data, etc. And batch _i_ would cover a different time period to batch _i+1_.
> 
> But, if I've understood correctly, if `temporally_align_batches` is set to `True` then the code repeatedly uses a single datetime. But the list of `t0_datetimes` could be different for each modality? Is that right?

So currently all the data sources are aligned in time. But each example has a different `t0`. 

This extra bit of code, makes an option to have all the examples _also_ align in time. This is for making examples for every gsp at the same time ready for a forecast. 

But if this isnt clear, there I should add some text (probably similar to what i just wrote) in the code

JackKelly:
Ah, got it, thank you for explaining!

> But if this isnt clear, there I should add some text (probably similar to what i just wrote) in the code

That would be great, thank you!  Would it be worth considering renaming `temporally_align_batches` to `temporally_align_all_examples` or something like that, just to be super-explicit?  (If I've understood correctly?

nowcasting_dataset/data_sources/fake/utils.py
peterdudfield: thanks, ive added something to stop thjis happening

nowcasting_dataset/data_sources/fake/utils.py
peterdudfield: Good shout using the `size=1`

Yea, I may leave it as a todo, if thats ok. Ill add a task for it

PR #599, opened 2022-02-01
add id to metadata

### LINKED ISSUE #598: Add id to metadata
## Detailed Description
Add id to metadata, this could hold the gsp, pv system id number. This would be useful for the 'forecast' service

## Possible Implementation
Add it to [pydantic model](https://github.com/openclimatefix/nowcasting_dataset/blob/main/nowcasting_dataset/data_sources/metadata/metadata_model.py)

peterdudfield:
# Pull Request

## Description

Add id to metadata, this is optional so can be None. 

Fixes #598 #600 

## How Has This Been Tested?

normal unittests

- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/consts.py
+++ b/nowcasting_dataset/consts.py
@@ -133,11 +133,5 @@ T0_DT = "t0_dt"
 SPATIAL_AND_TEMPORAL_LOCATIONS_OF_EACH_EXAMPLE_FILENAME = (
     "spatial_and_temporal_locations_of_each_example.csv"
 )
-SPATIAL_AND_TEMPORAL_LOCATIONS_COLUMN_NAMES = (
-    "t0_datetime_UTC",
-    "x_center_OSGB",
-    "y_center_OSGB",
-    "id",
-)
 
 LOG_LEVELS = ("DEBUG", "INFO", "WARNING", "ERROR")
--- a/nowcasting_dataset/consts.py
+++ b/nowcasting_dataset/consts.py
@@ -133,6 +133,11 @@ T0_DT = "t0_dt"
 SPATIAL_AND_TEMPORAL_LOCATIONS_OF_EACH_EXAMPLE_FILENAME = (
     "spatial_and_temporal_locations_of_each_example.csv"
 )
-SPATIAL_AND_TEMPORAL_LOCATIONS_COLUMN_NAMES = ("t0_datetime_UTC", "x_center_OSGB", "y_center_OSGB")
+SPATIAL_AND_TEMPORAL_LOCATIONS_COLUMN_NAMES = (
+    "t0_datetime_UTC",
+    "x_center_OSGB",
+    "y_center_OSGB",
+    "id",
+)
 
 LOG_LEVELS = ("DEBUG", "INFO", "WARNING", "ERROR")
--- a/nowcasting_dataset/data_sources/data_source.py
+++ b/nowcasting_dataset/data_sources/data_source.py
@@ -230,7 +230,11 @@ class DataSource:
         Get Batch Data
 
         Args:
-            locations: # TODO
+            locations: List of locations object
+                A location object contains
+                - a timestamp of the example (t0_datetime_utc),
+                - the x center location of the example (x_location_osgb)
+                - the y center location of the example(y_location_osgb)
 
         Returns: Batch data.
         """
@@ -420,12 +424,10 @@ class ZarrDataSource(ImageDataSource):
         Get Example data
 
         Args:
-            location: #TODO
-            # t0_datetime_utc: list of timestamps for the datetime of the batches.
-            #     The batch will also include data for historic and future depending
-            #     on `history_minutes` and `future_minutes`.
-            # x_center_osgb: x center batch locations
-            # y_center_osgb: y center batch locations
+            location: A location object of the example which contains
+                - a timestamp of the example (t0_datetime_utc),
+                - the x center location of the example (x_location_osgb)
+                - the y center location of the example(y_location_osgb)
 
         Returns: Example Data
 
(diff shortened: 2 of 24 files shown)


JackKelly (commented):
This looks good.

One thought is:

In the near future (this month?), we plan to create batches across the entire geospatial extent of the satellite data (over the ocean; over Europe; over Africa; etc.). (These training examples won't have PV data, of course. Instead they'll be used to pre-train the model to predict the next few frames of satellite imagery).

And, separately, I also plan to create batches which are centered on individual PV systems (and where the model will know nothing about GSPs).

And we plan to produce start using PV data from mainland Europe and elsewhere.

In all three of these scenarios, there won't be a "GSP ID". As such, is the approach proposed in this PR is compatible with our plans to create batches which aren't associated with a GSP? When there isn't a GSP then I suppose we could just fill the `id` columns with NaNs or something like that, but that maybe feels like a bit of a hack?

Also, if we do stick with including the GSP ID, then I'd advocate for changing the name to `gsp_id` rather than just `id`, just to be really explicit?

peterdudfield:
> This looks good.
> 
> One thought is:
> 
> In the near future (this month?), we plan to create batches across the entire geospatial extent of the satellite data (over the ocean; over Europe; over Africa; etc.). (These training examples won't have PV data, of course. Instead they'll be used to pre-train the model to predict the next few frames of satellite imagery).
> 
> And, separately, I also plan to create batches which are centered on individual PV systems (and where the model will know nothing about GSPs).
> 
> And we plan to produce start using PV data from mainland Europe and elsewhere.
> 
> In all three of these scenarios, there won't be a "GSP ID". As such, is the approach proposed in this PR is compatible with our plans to create batches which aren't associated with a GSP? When there isn't a GSP then I suppose we could just fill the `id` columns with NaNs or something like that, but that maybe feels like a bit of a hack?
> 
> Also, if we do stick with including the GSP ID, then I'd advocate for changing the name to `gsp_id` rather than just `id`, just to be really explicit?

Yea I agree it feels a bit of a hack. But yea that would be an option i.e set id to NaN or None. 

This is all work to get the forecast MVP working. Its because there is no GSP data to load, but I think we do need the GSP id for the MVP. One option would be to have some static GSP data, and do a hack when making batches/examples. 

Perhaps doing this: https://github.com/openclimatefix/nowcasting_dataset/issues/600, would help both tasks. Perhaps Ill give this a go, and itll actually make things tidier

peterdudfield:
Would be something like this, and then `Location` can be used in other general places in the code

```
class Location(BaseModel):
    """ Location of the example"""
    
    t0_datetime_utc: pd.Timestamp = Field(
        ...,
        description="The t0 of one example ",
    )

    x_center_osgb: float = Field(
        ...,
        description="The x center of one example in OSGB coordinates",
    )

    y_center_osgb: float = Field(
        ...,
        description="The y center of one example in OSGB coordinates",
    )

    id: Optional[int] = Field(
        None,
        description="The id of the GSP or the PV system. "
        "This is optional so can be None",
    )


class Metadata(BaseModel):
    """Class to store metadata data"""

    batch_size: int = Field(
        ...,
        g=0,
        description="The size of this batch. If the batch size is 0, "
        "then this item stores one data item",
    )

    locations: List[Location]

 ```

**Comment by JackKelly** [comment]
Cool beans.

But, um, isn't the GSP ID (and PV system ID) already in each batch? I haven't checked the code recently but I definitely remember feeding both the GSP ID and PV system ID into my ML models back in December :slightly_smiling_face:  Or am I getting confused?

**Comment by peterdudfield** [comment]
> Cool beans.
> 
> But, um, isn't the GSP ID (and PV system ID) already in each batch? I haven't checked the code recently but I definitely remember feeding both the GSP ID and PV system ID into my ML models back in December 🙂 Or am I getting confused?

No your right! Just that means for the live models we would need the data sources: gsp in order for the system to work. Even though the model might not use gsp data. Im making good progress with #600 so hopefully itll all come together

File: nowcasting_dataset/consts.py [hunk 1]
```diff
@@ -133,6 +133,11 @@
 SPATIAL_AND_TEMPORAL_LOCATIONS_OF_EACH_EXAMPLE_FILENAME = (
     "spatial_and_temporal_locations_of_each_example.csv"
 )
-SPATIAL_AND_TEMPORAL_LOCATIONS_COLUMN_NAMES = ("t0_datetime_UTC", "x_center_OSGB", "y_center_OSGB")
+SPATIAL_AND_TEMPORAL_LOCATIONS_COLUMN_NAMES = (
```
peterdudfield, line 136: remove this

nowcasting_dataset/data_sources/gsp/gsp_data_source.py
```diff
@@ -146,6 +145,7 @@ def get_all_locations(
             1. list of datetimes
             2. list of x locations
             3. list of y locations
+            4. list of gsp ids
```
peterdudfield, line 148: update return description

peterdudfield:
Comments for chat with @JackKelly and @jacobbieker 
-change `Location` to `SpaceTimeLocation`
- add id_type, to show what type of id it is

nowcasting_dataset/data_sources/data_source.py
```diff
@@ -378,9 +347,7 @@ def _get_time_slice(self, t0_datetime_utc: pd.Timestamp):
 
     def get_example(
         self,
-        t0_datetime_utc: pd.Timestamp,  #: Datetime of "now": The most recent obs.
-        x_center_osgb: Number,  #: Centre, in OSGB coordinates.
-        y_center_osgb: Number,  #: Centre, in OSGB coordinates.
+        location: SpaceTimeLocation,  #: Datetime of "now": The most recent obs.
```
jacobbieker, line 350:
```suggestion
        location: SpaceTimeLocation,  #: Location object of the most recent observation
```

jacobbieker (approved):
Looks good to me! Just one small comment. Good job!

peterdudfield:
@JackKelly im gona merge this, to push on the MVP, but please free feel to add comments, and ill do those changes

PR #660, opened 2022-05-16
Load data into memory if live

jacobbieker:
# Pull Request

## Description

Loads the Zarr arrays into memory if it is being run as live, as otherwise the Zarr store might disappear in the middle from being updated by the 5 minutely process.

Fixes https://github.com/openclimatefix/nowcasting_forecast/issues/68

## How Has This Been Tested?

- [ ] Yes

_If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

- [ ] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/satellite/satellite_data_source.py
+++ b/nowcasting_dataset/data_sources/satellite/satellite_data_source.py
@@ -107,12 +107,17 @@ class SatelliteDataSource(ZarrDataSource):
         ).transform
 
     def _open_data(self) -> xr.DataArray:
-        return open_sat_data(
+        data_array = open_sat_data(
             zarr_path=self.zarr_path,
             consolidated=self.consolidated,
             logger=self.logger,
             sample_period_minutes=self.sample_period_minutes,
         )
+        # If live, then load into memory so the data
+        # doesn't disappear while its being used
+        if self.is_live:
+            data_array = data_array.load()
+        return data_array
 
     @staticmethod
     def get_data_model_for_batch():


peterdudfield:
@all-contributors please add @jacobbieker  for review

JackKelly:
how bullet-proof is this? what happens if the Zarr store on disk disappears while it's being loaded into RAM? :slightly_smiling_face: Or if the chunks on disk change after the metadata has been loaded?

Should we consider saving new Zarrs to different Zarr filenames, and deleting them after, say, 6 hours? So we can guarantee that a Zarr won't disappear while it's being read (assuming it's being read less than 6 hours after the Zarr was created!)

jacobbieker:
> how bullet-proof is this? what happens if the Zarr store on disk disappears while it's being loaded into RAM? :slightly_smiling_face: Or if the chunks on disk change after the metadata has been loaded?
> 
> Should we consider saving new Zarrs to different Zarr filenames, and deleting them after, say, 6 hours? So we can guarantee that a Zarr won't disappear while it's being read (assuming it's being read less than 6 hours after the Zarr was created!)
> 
> 

Yeah, there is the issue of race conditions, we've tried to minimize it by writing to a different file then renaming after it's finished which is fairly quick. @peterdudfield and I are planning on making it more robust. Having different filenames does sound like a good alternative!

JackKelly:
> writing to a different file then renaming after it's finished

Great idea!! If renaming is an atomic operation on AWS then I would guess that's bullet-proof already? (but I guess that even if renaming a single file is atomic; renaming all the files in a Zarr is unlikely to be atomic?)

Awesome work, BTW! I'm super-impressed at home much stuff you guys are getting done!

jacobbieker:
> > writing to a different file then renaming after it's finished
> 
> Great idea!! If renaming is an atomic operation on AWS then I would guess that's bullet-proof already? (but I guess that even if renaming a single file is atomic; renaming all the files in a Zarr is unlikely to be atomic?)
> 
> Awesome work, BTW! I'm super-impressed at home much stuff you guys are getting done!

I don't know if it's atomic, I think it should be, but not sure. But each zarr is a single file, it's stored as a zip file like I've been doing for HF datasets and that seems to be working quite well.

JackKelly:
nice! According to [this](https://stackoverflow.com/a/69925633/732596), moving a single file is atomic on AWS. So I _think_ that means that what you've already build should be bullet-proof. Nice!

If it's not then another option might be a good-old-fashioned LOCKFILE. But let's assume your current implementation is bullet-proof :slightly_smiling_face:

PR #667, opened 2022-05-19
pv live interpolation options

peterdudfield:
# Pull Request

## Description

Ad option to load extra live data and interpolate forward

Fixes #https://github.com/openclimatefix/nowcasting_forecast/issues/95

## How Has This Been Tested?

added a extra test first

- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/pv/live.py
+++ b/nowcasting_dataset/data_sources/pv/live.py
@@ -51,13 +51,13 @@ def get_pv_power_from_database(
     """
     Get pv power from database
 
-    :param history_duration: a timedelta of how many minutes to load in the past
-    :param interpolate_minutes: how many minutes we should interpolate the data froward for
-    :param load_extra_minutes: the extra minutes we should load, in order to load more data.
-        This is because some data from a site lags significantly behind 'now'
+    Args:
+        history_duration: a timedelta of how many minutes to load in the past
+        interpolate_minutes: how many minutes we should interpolate the data froward for
+        load_extra_minutes: the extra minutes we should load, in order to load more data.
+            This is because some data from a site lags significantly behind 'now'
 
-    Returns: pandas data frame with the following columns
-    - pv systems indexes
+    Returns:pandas data frame with the following columns pv systems indexes
     The index is the datetime
 
     """
--- a/nowcasting_dataset/data_sources/pv/live.py
+++ b/nowcasting_dataset/data_sources/pv/live.py
@@ -45,9 +45,16 @@ def get_metadata_from_database() -> pd.DataFrame:
     return pv_systems_df
 
 
-def get_pv_power_from_database(history_duration: timedelta) -> pd.DataFrame:
+def get_pv_power_from_database(
+    history_duration: timedelta, interpolate_minutes: int = 30, load_extra_minutes: int = 60
+) -> pd.DataFrame:
     """
     Get pv power from database
+
+    :param history_duration: a timedelta of how many minutes to load in the past
+    :param interpolate_minutes: how many minutes we should interpolate the data froward for
+    :param load_extra_minutes: the extra minutes we should load, in order to load more data.
+        This is because some data from a site lags significantly behind 'now'
 
     Returns: pandas data frame with the following columns
     - pv systems indexes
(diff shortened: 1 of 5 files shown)


nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -45,23 +45,42 @@ def get_metadata_from_database() -> pd.DataFrame:
...
 
 
-def get_pv_power_from_database(history_duration: timedelta) -> pd.DataFrame:
+def get_pv_power_from_database(
+    history_duration: timedelta, interpolate_minutes: int = 30, load_extra_minutes: int = 60
+) -> pd.DataFrame:
     """
     Get pv power from database
 
+    :param history_duration: a timedelta of how many minutes to load in the past
```
JackKelly, line 54: Tiny nit: I think we're using numpy-style docstring in `nowcasting_dataset`? https://numpydoc.readthedocs.io/en/latest/format.html#parameters

nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -45,23 +45,42 @@ def get_metadata_from_database() -> pd.DataFrame:
     return pv_systems_df
 
 
-def get_pv_power_from_database(history_duration: timedelta) -> pd.DataFrame:
+def get_pv_power_from_database(
+    history_duration: timedelta, interpolate_minutes: int = 30, load_extra_minutes: int = 60
```
JackKelly, line 49: I'd recommend settings defaults in exactly one place: either the `model.py` or here. In the past, I've found it slightly hard to debug code when defaults get set in multiple places :slightly_smiling_face:

nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -45,23 +45,42 @@ def get_metadata_from_database() -> pd.DataFrame:
...
     # make database connection
     url = os.getenv("DB_URL_PV")
     db_connection = DatabaseConnection(url=url, base=Base_PV)
 
     with db_connection.get_session() as session:
-        start_utc = datetime.now(tz=timezone.utc) - history_duration
-        pv_yields: List[PVYieldSQL] = get_pv_yield(session=session, start_utc=start_utc)
+        pv_yields: List[PVYieldSQL] = get_pv_yield(session=session, start_utc=start_utc_extra)
+
+        logger.debug(f"Found {len(pv_yields)} from database")
```
JackKelly, line 83:
```suggestion
        logger.debug(f"Found {len(pv_yields)} PV systems from database")
```

nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -90,7 +110,16 @@ def get_pv_power_from_database(history_duration: timedelta) -> pd.DataFrame:
 
     pv_yields_df.columns = encode_label(pv_yields_df.columns, label="pvoutput")
 
-    # interpolate in between, maximum 30 mins
-    pv_yields_df.interpolate(limit=3, limit_area="inside", inplace=True)
+    # interpolate in between, maximum 'live_interpolate_minutes' mins
+    # note data is in 5 minutes chunks
+    pv_yields_df = empty_df.join(pv_yields_df)
```
JackKelly, line 115: Instead of creating an `empty_df` and `join`ing it, would something like this be a valid alternative:

```python
# Append a row of NaNs at `now`:
pv_yields_df = pv_yields_df.append(pd.Series(name=now))

# Interpolate up until `now`:
pv_yields_df = pv_yields_df.resample("5T").interpolate()
```

JackKelly (approved):
LGTM! A few small suggestions! :slightly_smiling_face:

nowcasting_dataset/data_sources/pv/live.py
peterdudfield: Thanks

nowcasting_dataset/data_sources/pv/live.py
peterdudfield: ok, I can remove them here

nowcasting_dataset/data_sources/pv/live.py
peterdudfield: thanks

nowcasting_dataset/data_sources/pv/live.py
peterdudfield: I like the idea, but i think its safer with the join. It covers things like
1. what if the dataframe has no values in it?
2. What if there are no values for the start time, we should then add one at the start then
3. What if the now is already in the df?
4. What if we dont want to interpolate the values for more than 1 hour?

peterdudfield: Im gona merge, and we can always come back to it

PR #678, opened 2022-05-31
Gsp cubic interpolation

peterdudfield:
# Pull Request

## Description

- use cubic interpolation
- floor time, dont use ceil

Fixes #https://github.com/openclimatefix/nowcasting_forecast/issues/114

## How Has This Been Tested?

- [ ] Yes

_If your changes affect data processing, have you plotted any changes? i.e. have you done a quick sanity check?_

- [ ] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/time.py
+++ b/nowcasting_dataset/time.py
@@ -296,3 +296,23 @@ def make_random_time_vectors(
         "time_30": utils.to_numpy(time_30.T).astype(np.int32),
         "time_60": utils.to_numpy(time_60.T).astype(np.int32),
     }
+
+
+def floor_minutes_dt(dt, minutes: Optional[int] = 30):
+    """
+    Floor a datetime by 30 mins.
+
+    For example:
+    2021-01-01 17:01:01 --> 2021-01-01 17:00:00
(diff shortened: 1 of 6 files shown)


nowcasting_dataset/time.py
```diff
@@ -296,3 +296,23 @@ def make_random_time_vectors(
         "time_30": utils.to_numpy(time_30.T).astype(np.int32),
         "time_60": utils.to_numpy(time_60.T).astype(np.int32),
     }
+
+
+def floor_minutes_dt(dt, minutes: Optional[int] = 30):
```
JackKelly, line 301: oooh, the little I know about writing functions to process datetimes suggests that there are enough fiddly edge-cases to mean that it's almost always better to use existing, well-tested libraries for processing datetimes, like Pandas! :slightly_smiling_face: . So I'd be very tempted to replace this custom function with `return pd.Timestamp(dt).floor(pd.Timedelta(minutes=minutes))`

peterdudfield: I agree, just sometimes feels like an over kill to use pandas when we are in 'datetime' world. But yea, I think yoru right in this case

PR #679, opened 2022-06-01
Pv interpolation

peterdudfield:
# Pull Request

## Description

-use cubic interpolation
- restrict PV data to necessary time

Fixes #

## How Has This Been Tested?

added to tests

- [ ] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/pv/live.py
+++ b/nowcasting_dataset/data_sources/pv/live.py
@@ -2,7 +2,7 @@
 import logging
 import os
 from datetime import datetime, timedelta, timezone
-from typing import List
+from typing import List, Optional
 
 import pandas as pd
 from nowcasting_datamodel.connection import DatabaseConnection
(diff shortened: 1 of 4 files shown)


nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -120,7 +125,12 @@ def get_pv_power_from_database(
     pv_yields_df = empty_df.join(pv_yields_df)
     limit = int(interpolate_minutes / 5)
     if limit > 0:
-        pv_yields_df.interpolate(limit=limit, inplace=True)
+        try:
+            pv_yields_df.interpolate(limit=limit, inplace=True, method="cubic")
+        except Exception as e:
```
JackKelly, line 130: ooh, it's dangerous to catch all exceptions! What happens if `df.interpolate` raises a _different_ exception? That different exception will just get silently swallowed!

I'd suggest we either:

* catch the precise exception subclass we're interested in, or
* add a message to the exception, and re-raise the original exception, like [this](https://stackoverflow.com/a/46091127/732596)

nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -120,7 +125,12 @@ def get_pv_power_from_database(
...
+            logger.exception(e)
+            logger.debug(pv_yields_df)
```
JackKelly, line 132: also, I think this might fail, because `logger.debug` expects a string not a df?

peterdudfield: no this is fine, it prints out the same as if you did print(df)

nowcasting_dataset/data_sources/pv/live.py
peterdudfield: ill put it in here https://github.com/openclimatefix/nowcasting_dataset/issues/681

PR #700, opened 2022-06-28
2D osgb sat coords

### LINKED ISSUE #695: Fake `hrvsatellite` and `satellite` batches need 2D osgb coords
At the moment, `nowcasting_dataset.data_sources.fake.batch.create_image_array` creates fake satellite data with 1D `x_geostationary` and `y_geostationary` coords. We also need 2D `x_osgb` and `y_osgb` coords.

I just wanted to check with you, @peterdudfield, that this change won't break too many things?!

JackKelly:
# Pull Request

This is the correct PR for this issue!

## Description

Give fake hrv batches 2D osgb coords. Because "real" satellite data has 2D osgb coords.

Fixes #695

## How Has This Been Tested?

- [x] Yes, unittests pass

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/fake/batch.py
+++ b/nowcasting_dataset/data_sources/fake/batch.py
@@ -548,12 +548,16 @@ def create_image_array(
     if t0_datetime_utc is None:
         t0_datetime_utc = make_t0_datetimes_utc(batch_size=1)[0]
 
+    # We want the OSGB coords to be 2D for satellite data:
+    two_dimensional_osgb_coords = nwp_or_satellite == "satellite"
+
+    # Get OSB coords:
     x_osgb, y_osgb = make_image_coords_osgb(
         size_y=image_size_pixels_height,
         size_x=image_size_pixels_width,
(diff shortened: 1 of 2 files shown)


nowcasting_dataset/data_sources/fake/batch.py
```diff
@@ -553,39 +553,41 @@ def create_image_array(
         size_x=image_size_pixels_width,
         x_center_osgb=x_center_osgb,
         y_center_osgb=y_center_osgb,
+        two_dimensional=nwp_or_satellite == "satellite",
```
peterdudfield, line 556: what about hrv?

peterdudfield: maybe we dont need to worry about that

peterdudfield: Perhaps it would be better to do it as 
```
two_dimensional = Optional[bool] = None

if two_dimensional is None:
      two_dimensional=nwp_or_satellite == "satellite"
```

Just seems a bit confusing in the function inputs

PR #702, opened 2022-06-29
return array of nans when there is no pv data

peterdudfield:
# Pull Request

## Description

Return array of nans instead of no data, when there are no pv yields
The pv systems is filled with 0, if elevation is below -10. This helps for live data, so that not all the data is removed, when removing nans

Fixes https://github.com/openclimatefix/nowcasting_forecast/issues/122

## How Has This Been Tested?

normal unitest tests, + extra ones

- [x] Yes

## Checklist:

### DIFF
--- a/nowcasting_dataset/data_sources/pv/live.py
+++ b/nowcasting_dataset/data_sources/pv/live.py
@@ -43,6 +43,8 @@ def get_metadata_from_database(providers: List[str] = None) -> pd.DataFrame:
     pv_system_all_df = []
     for provider in providers:
 
+        logger.debug(f"Get PV systems from database for {provider}")
+
         with db_connection.get_session() as session:
             # read pv systems
             pv_systems: List[PVSystemSQL] = get_pv_systems(session=session, provider=provider)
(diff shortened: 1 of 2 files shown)


nowcasting_dataset/data_sources/pv/live.py
```diff
@@ -53,17 +56,19 @@ def get_metadata_from_database(providers: List[str] = None) -> pd.DataFrame:
...
 
         if len(pv_systems_df) == 0:
-            return pd.DataFrame(
+            pv_systems_df = pd.DataFrame(
                 columns=["pv_system_id", "latitude", "longitude", "installed_capacity_kw"]
             )
-
-        pv_systems_df.index = encode_label(pv_systems_df["pv_system_id"], label=provider)
-        pv_systems_df = pv_systems_df[["latitude", "longitude", "installed_capacity_kw"]]
+        else:
```
peterdudfield, line 62: this solves different bug

jacobbieker (approved):
LGTM!

